PR #25, opened 2020-05-26
titles: fix error showing in TitlesField

zzacharo:


src/lib/DepositRecordSerializer.js
```diff
@@ -5,38 +5,38 @@
...
+      let pickedValues = _.pickBy(mappedValues, (value) => {
+        if (_.isArray(value) || _.isObject(value)) {
+          return !_.isEmpty(value);
+        }
+        return !_.isNull(value);
+      });
+      return pickedValues;
     }
-    return result;
+    return obj ? obj : null;
```
fenekku, line 35: So for `obj = 0`, we get `null` here which *might* give some weird results... 

I am just realizing that we might need to add a note about how this works for the PUT / POST approach. Once we attempt to use PATCH and only send changes it will have to be revisited.

zzacharo: Nice catch! Definitely we need to add tests on that with all possible cases! I will create an issue for that. In the meanwhile, I changed the line to:
```suggestion
    return _.isNumber(obj) || obj ? obj : null;
```

PR #40, opened 2020-06-10
lib: add identifiers for creators/contributors

### LINKED ISSUE #26: Test `DepositRecordSerializer`
- [x] Test `serialize` method
- [x] Test `deserialize` method
- [x] Test `removeEmptyValues` method

fenekku:
- serializes identifiers correctly
- setup test harness for DepositRecordSerializer

helps close: https://github.com/inveniosoftware/invenio-rdm-records/issues/97
closes #26

### DIFF
--- a/src/lib/DepositRecordSerializer.js
+++ b/src/lib/DepositRecordSerializer.js
@@ -14,13 +14,17 @@ import _pickBy from 'lodash/pickBy';
 import _mapValues from 'lodash/mapValues';
 
 export class DepositRecordSerializer {
-  constructor() {
-    this.removeEmptyValues = this.removeEmptyValues.bind(this);
-  }
+
   deserialize(record) {
     return record;
   }
 
+  /**
+  * Remove empty fields from record
+  * @method
+  * @param {object} obj - potentially empty object
+  * @returns {object} record - without empty fields
+  */
   removeEmptyValues(obj) {
     if (_isArray(obj)) {
       let mappedValues = obj.map((value) => this.removeEmptyValues(value));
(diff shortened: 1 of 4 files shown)


src/lib/DepositRecordSerializer.js
```diff
@@ -41,8 +45,69 @@ export class DepositRecordSerializer {
...
+  * Transform frontend creators structure to API-compatible structure.
+  *
+  * NOTE: Serialization doesn't deal with validation: safely access properties
+  * NOTE: If property absent from input, it should be absent from output
+  * @method
+  * @param {object} record - with creators in frontend format
+  * @returns {object} record - with creators in API format
+  */
+  serializeCreators(record) {
+    const in_creators = record.creators || [];
```
zzacharo, line 58: @fenekku why should we change the `identifiers` from an object in the backend to an array in the frontend? I think is better to be consistent with the fields' schema. Is there any particular reason that I miss?

src/lib/DepositRecordSerializer.test.js
```diff
@@ -0,0 +1,171 @@
...
+        null,
+        undefined,
+        {},
+      ],
+      description: ""
+    }
+
+    const cleanedRecord = serializer.removeEmptyValues(record);
+
+    expect(cleanedRecord).toEqual({version: 0});
```
zzacharo, line 30: Shouldn't it return `{ cool: false, version: 0}`?

src/lib/components/CreatorOrContributorField.js
```diff
@@ -74,7 +74,7 @@ const _CreatorOrContributorField = ({ field, form, ...props }) => {
     form.setFieldValue(nameFieldPath, `${givenName} ${familyName}`);
   };
 
-  const personValue = 'Personal';
+  const isPerson = getIn(form.values, typeFieldPath, null) === 'Personal'
```
zzacharo, line 77: I would rename it probably to `isPersonalType`. I think it's better to give the exact meaning of this variable, WDYT?

src/lib/DepositRecordSerializer.js
fenekku: We do this dance of backend object -> frontend array -> to backend object because the component used to fill the identifiers is an array component. The user can add many different identifiers and remove them: visually it's an array. We don't have a component that is visually like an array but backs to an object, so the ArrayField was used and it does the job :smile:

src/lib/DepositRecordSerializer.test.js
fenekku: The logic of `removeEmptyValues` wasn't altered, so you tell me :) . But seriously, I think you are right actually: if the `missing` value for that Boolean field on the backend is True, then when it is set to False on the frontend, the serialization would strip it and the backend not seeing it would set the field back to True... This is actually true for all fields... :worried: 

Let's discuss how we want to proceed on that one.

src/lib/components/CreatorOrContributorField.js
fenekku: Ah, I think in this case the *exact meaning* is that a creator/contributor is a person or an organization. The fact that we use "Personal" or "Organizational" as the *value* to indicate that is a detail: we could have used "person"/"organization" or "p"/"o" or any other symbol. And if we do change that symbol, we don't have to change the variable name because it would still reflect the meaning. It's then easier to read: the abstraction makes the code easier to understand while not lying.

zzacharo: Ah ok, I took the meaning of the value for granted :) Ok then it's ok with me!

src/lib/DepositRecordSerializer.js
zzacharo: After our discussion, I created a generic issue for this matter https://github.com/inveniosoftware/react-invenio-deposit/issues/42

src/lib/DepositRecordSerializer.test.js
zzacharo: @fenekku can you create an issue for that? Then, you can go ahead and merge this/release :)

src/lib/DepositRecordSerializer.js
fenekku: :+1:

src/lib/DepositRecordSerializer.test.js
fenekku: :+1: Created: https://github.com/inveniosoftware/react-invenio-deposit/issues/43

Issue #42, opened 2020-06-25
serializers: keep to minimal the transformation between REST Api and UI


zzacharo:
We should feed the UI components as much as we can with the same objects/fields (i.e schema) that we receive from the API response. This will make easier the understanding of the data format that the components are expecting and might decomplexify the debugging of issues in the future.

fenekku:
As much as we can, we should. Identifiers in the backend are stored in a dict and formik only allows us to manipulate arrays, so there will be some discrepancies like this.

PR #44, opened 2020-06-26
lib: clear empty default contributors

fenekku:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/195

### DIFF
--- a/src/lib/DepositRecordSerializer.js
+++ b/src/lib/DepositRecordSerializer.js
@@ -7,6 +7,7 @@
 
 import _isNumber from 'lodash/isNumber';
 import _isEmpty from 'lodash/isEmpty';
+import _isEqual from 'lodash/isEqual';
 import _isObject from 'lodash/isObject';
 import _isArray from 'lodash/isArray';
 import _isNull from 'lodash/isNull';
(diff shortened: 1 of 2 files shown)


src/lib/DepositRecordSerializer.js
```diff
@@ -79,7 +80,19 @@ export class DepositRecordSerializer {
   */
   serializeContributors(record) {
     const in_contributors = record.contributors || [];
-    const contributors = in_contributors.map((contributor) => {
+
+    // Remove contributors with only a type
+    // Note: we have to do this because type is filled by default, but
+    // contributors is an optional field
+    let contributors = in_contributors.filter((contributor) => {
```
fenekku, line 87: This is more of what you mentioned @zzacharo :disappointed: but I don't know of a good alternative for default values. 

We are in this situation where we want a default value to be shown AND stored IF other fields of the contributor components are filled... Otherwise we would still need to do fiddly serialization work where we inject the default value if contributors has other values which is the reverse  of what we doing here: removing the default value if contributors has no other values  

Other ideas? Maybe Formik has a way to deal with that kind of problem?

fenekku:
Rebase and merging per @zzacharo agreement in chat. The whole deserialization will be considered here: https://github.com/inveniosoftware/react-invenio-deposit/issues/42

PR #55, opened 2020-09-16
files: add file uploader for deposit form

zzacharo:
Addresses https://github.com/inveniosoftware/invenio-app-rdm/issues/270

Screenshot 2020-09-16 at 15 44 38

- Styling needs more work
- The uploader has a hardcoded bucket id at the moment. @lnielsen should I create one, if there is no for the record? Or we shelve it for when we finalize the backend?


src/lib/DepositController.js
```diff
@@ -9,29 +9,38 @@
...
+} from './state/types';
 
 export class DepositController {
-  constructor(apiClient) {
+  constructor(apiClient, fileUploader) {
     this.apiClient = apiClient;
+    this.fileUploader = fileUploader;
   }
 
   exists(record) {
```
ntarocco, line 24: I would rename it to something more clear: `draftPreviouslyCreated` for example, it is not very clear when you use it otherwise: `if this.exist()...`

src/lib/DepositController.js
```diff
@@ -9,29 +9,38 @@
...
 
   validate(record) {
     console.log('Validate record', record);
   }
 
-  async save_draft(record, { formik, store }) {
+  async saveDraft(record, { formik, store }) {
     // Saves a draft of the record
     const recordSerializer = store.config.recordSerializer;
     let payload = recordSerializer.serialize(record);
```
ntarocco, line 35:
```suggestion
    const payload = recordSerializer.serialize(record);
```

src/lib/DepositController.js
```diff
@@ -68,11 +81,50 @@ export class DepositController {
...
+    try {
+      const resp = await this.apiClient.deleteFile(fileVersionUrl);
+      store.dispatch({
+        type: 'FILE_DELETED_SUCCESS',
+        payload: {
+          fileName: file.fileName,
+        },
+      });
+    } catch (e) {
+      console.log('error');
```
ntarocco, line 126: I guess error reporting to user will be improved...

src/lib/DepositFileUploader.js
```diff
@@ -0,0 +1,123 @@
...
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+// Drives the business logic of the InvenioFormApp.
+// Defines what happens when a button is clicked.
+
+import axios from 'axios';
+import _indexOf from 'lodash/indexOf';
+
+export class DepositFileUploader {
```
ntarocco, line 14: silly question: do we still call all this `Deposit` or not it `Draft`?

src/lib/components/FileManager/FileManager.js
```diff
@@ -0,0 +1,78 @@
...
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Dropzone from 'react-dropzone';
+import { FileArea } from './FileArea';
+import { UploadState } from '../../state/reducers/files';
+
+export default class FileManager extends Component {
```
ntarocco, line 13: another silly one: it sounds weird to have a cmp called "FileManager" :) Is it more about the component that renders the uploaded files right and the dropzone right?

src/lib/store.js
```diff
@@ -11,11 +11,20 @@ import thunk from 'redux-thunk';
 import rootReducer from './state/reducers';
 import { INITIAL_STORE_STATE } from './storeConfig';
 
+const preloadFiles = (files) =>
+  files
+    .map((file) => ({ state: 'finished', ...file }))
+    .reduce((acc, current) => {
+      acc[current.filename] = { ...current };
```
ntarocco, line 18: I am not sure what you are trying to do here: isn't another `map` rather than a `reduce`?

src/lib/DepositController.js
zzacharo: I renamed it to `draftAlreadyCreated` :)

src/lib/DepositController.js
zzacharo: It's a `let` because in case of creation I assign the request `response.data` to it!

src/lib/DepositController.js
zzacharo: Yes it's far from finished that part :)

src/lib/DepositFileUploader.js
zzacharo: We kept the deposit as a concept and we use draft for the product that the deposit is managing. A deposit is creating/saving/publishing drafts!

src/lib/components/FileManager/FileManager.js
zzacharo: Did some refactoring on the name and merged `FileManager` and `FileArea` in one `FileUploader` file. What do you think?

src/lib/store.js
zzacharo: `files` coming from backend are a list and with the `reduce` I am converting them into an object with key the filename and value the object found in the list. E.g

```javascript
// Backend files
files = [{filename: 'Foo.txt', size: 1000, ...}

// Redux files strucutre
files = {
  'Foo.txt': {
    size: 1000,
    ....
  }
}

```

ntarocco (approved):

PR #56, opened 2020-10-05
errors: improve frontend handling of errors

fenekku:
**EDIT**: See comment about it not being rebased. I leave it up to you to decide.

work in progress of frontend errors. Here is a screenshot of how things look like currently:

oct5_upload_errors

Co need:
- https://github.com/inveniosoftware/invenio-app-rdm/pull/293
- https://github.com/inveniosoftware/invenio-rdm-records/pull/182
 
To note:
- I replaced "family name" / "given name" by "Name" because the backend does validation against name and it's simpler for now to get both person and organization dealt with in 1 swoop.

Things that need to be done:
- Not send a contributor that is empty except for `type` by default: this makes the backend validate it and return an error on the frontend.
- Convert machine values to human labels for access radio button. I think this conversion should take place on the frontend. Translation here will be a challenge.
- Deal with resource type error correctly

### DIFF
--- a/src/lib/state/actions/deposit.js
+++ b/src/lib/state/actions/deposit.js
@@ -7,10 +7,16 @@
 
 import { FORM_ACTION_FAILED, FORM_ACTION_EVENT_EMITTED } from '../types';
 
-export const setFormErrorsFromResponse = (response, formik) => {
+
+/**
+ * Closure over Axios error and formik, returning async error setting function.
+ *
+ * @param {Error} error - Axios error
+ * @param {Formik state} formik
+ */
+export const setFormErrors = (error, formik, record) => {
   return async (dispatch, getState, config) => {
-    const errorHandler = config.apiErrorHandler;
-    const extractedErrors = errorHandler.extractErrors(response);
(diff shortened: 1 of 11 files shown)


fenekku:
Yes this is not rebased on master, but we are not supporting file upload for October 8. My take is we branch master off to keep Zach's work and then `git reset --hard` to this branch.

src/lib/state/actions/deposit.js
```diff
@@ -7,10 +7,16 @@
...
+ * Closure over Axios error and formik, returning async error setting function.
+ *
+ * @param {Error} error - Axios error
+ * @param {Formik state} formik
+ */
+export const setFormErrors = (error, formik, record) => {
   return async (dispatch, getState, config) => {
-    const errorHandler = config.apiErrorHandler;
-    const extractedErrors = errorHandler.extractErrors(response);
+    const extractedErrors = config.apiErrorHandler.extractErrors(error, record);
```
zzacharo, line 19: @fenekku why do we need the record?

zzacharo (approved):

PR #73, opened 2020-10-30
List of components

### LINKED ISSUE #9: Add LanguagesField
See #4 

Data Model Question:

Can this field be repeatable? If the record contains multiple languages the https://en.wikipedia.org/wiki/ISO_639-3 code `mul` can be used but it doesn't specify which languages. This came up from librarians on our side.

**EDIT**
Yes, multiple languages are supported.

### LINKED ISSUE #15: Add Related Identifiers (related work)
See #4 . ~~They are similar to each other and in the same section so might as well do both at the same time.~~ Splitting them.

### LINKED ISSUE #17: DOI/Alternative Identifier Fields
See #4 (grouped functionality so together)

zzacharo:
Adding the below components
- Languages; closes #9 
- Alternate Identifiers; closes #17  
- Related Identifiers; closes #15 

Screenshot 2020-10-30 at 15 23 52
Screenshot 2020-10-30 at 15 24 03

### DIFF
--- a/src/lib/components/LanguagesField.js
+++ b/src/lib/components/LanguagesField.js
@@ -1,0 +1,48 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2020 CERN.
+// Copyright (C) 2020 Northwestern University.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { FieldLabel, SelectField } from 'react-invenio-forms';
+
+export class LanguagesField extends Component {
+  render() {
+    const { fieldPath, label, labelIcon, options, required } = this.props;
+    return (
+      <SelectField
+        fieldPath={fieldPath}
+        label={
+          <FieldLabel htmlFor={fieldPath} icon={labelIcon} label={label} />
+        }
+        options={options}
+        placeholder={'Select languages...'}
+        multiple
+        search
+        required={required}
+      />
+    );
+  }
+}
+
+LanguagesField.propTypes = {
+  fieldPath: PropTypes.string.isRequired,
+  label: PropTypes.string,
+  labelIcon: PropTypes.string,
+  options: PropTypes.arrayOf(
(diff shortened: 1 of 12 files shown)


src/lib/components/LanguagesField.js
```diff
@@ -0,0 +1,48 @@
...
+        label={
+          <FieldLabel htmlFor={fieldPath} icon={labelIcon} label={label} />
+        }
+        options={options}
+        placeholder={'Select languages...'}
+        multiple
+        search
+        required={required}
+      />
+    );
```
fenekku, line 27: Let's use an array of SelectFields since we support multiple languages in the latest schema changes.

src/lib/fields/LanguagesField.js
```diff
@@ -0,0 +1,28 @@
...
+import _cloneDeep from 'lodash/cloneDeep';
+
+import { Field } from './Field';
+
+export class LanguagesField extends Field {
+  deserialize(record) {
+    let langu = _get(record, this.fieldpath, this.deserializedDefault);
+    identifiers = identifiers.map((identifier) => ({}));
+    return record;
+  }
```
fenekku, line 19: I think something is not quite right here :grin: ...

fenekku (commented):
I am sorry, there is something with the LanguagesField :sweat_smile:

src/lib/fields/LanguagesField.js
zzacharo: Oups, this is a leftover! You can delete it!

src/lib/components/LanguagesField.js
zzacharo: The languages field uses the multiple prop from the dripdown component so you can select multiple languages... Do you think this isn't enough?

fenekku: Ah! My bad I didn't see that. That should work then.

src/lib/fields/LanguagesField.js
fenekku: Got it. I will take over this PR :+1:

fenekku:
Related work update:

image


Will look even better when resource type is changed into a single dropdown field.

Issue #11, opened 2020-05-15
Add LicenseField


fenekku:
See #4 

PRs:

* license vocabulary (inveniosoftware/invenio-rdm-records#279)
* create LicenseField in react-invenio-deposit (#90)
* add LicenseField to invenio-app-rdm's RDMDepositForm (pending)

fenekku:
Safe to say it was added in #128 . There are still outstanding issues but those have been reported separately: e.g. https://github.com/inveniosoftware/react-invenio-deposit/issues/152

PR #90, opened 2020-11-13
Add LicenseField

alexdutton:
This doesn't yet properly update the record, but does show the options correctly.

I've added a search prop to the DropDown as there are so many options in the vocabulary.

It doesn't use SelectField, as there's not the same mapping between what is selected, and what needs to be inserted into the record.

Part of #11.

### DIFF
--- a/src/lib/components/LicenseField.js
+++ b/src/lib/components/LicenseField.js
@@ -1,0 +1,58 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2020 CERN.
+// Copyright (C) 2020 Northwestern University.
+// Copyright (C) 2020 Cottage Labs LLP.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React, { Component, useMemo } from 'react';
+import PropTypes from 'prop-types';
+
+import { FieldLabel, GroupField, SelectField, TextField } from 'react-invenio-forms';
+import { Field, getIn } from 'formik';
+import { Form } from 'semantic-ui-react';
+
+export function LicenseField({ vocabulary, fieldPath }) {
+  let itemKey = ({ scheme, identifier }) => `${scheme}:${identifier}`;
+
+  let vocabularyByValue = {};
+  vocabulary.forEach(item => {
+    vocabularyByValue[itemKey(item)] = item;
+  });
+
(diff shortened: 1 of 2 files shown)


src/lib/components/LicenseField.js
```diff
@@ -0,0 +1,58 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2020 CERN.
+// Copyright (C) 2020 Northwestern University.
+// Copyright (C) 2020 Cottage Labs LLP.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import React, { Component, useMemo } from 'react';
```
fenekku, line 9: can remove `useMemo` and other unused imports

src/lib/components/LicenseField.js
```diff
@@ -0,0 +1,58 @@
...
+import PropTypes from 'prop-types';
+
+import { FieldLabel, GroupField, SelectField, TextField } from 'react-invenio-forms';
+import { Field, getIn } from 'formik';
+import { Form } from 'semantic-ui-react';
+
+export function LicenseField({ vocabulary, fieldPath }) {
```
fenekku, line 16: (These are more general overall comments before you get too deep)

Since the creation of this issue and now, there were a couple of changes to the [metadata schema](https://github.com/inveniosoftware/docs-invenio-rdm/blob/master/docs/reference/metadata.md#rights-0-n):

- perhaps `RightsField` is a better name now
- There can be multiple licenses selected

We've been using the class-based component approach, so probably best to stick with that (or convince us otherwise to change the rest! :grin:  ).

You can use the `FundingField` as a template for this field. It also had to deal with how the `SelectField` doesn't store what Formik stores in the form. It's a decent example of how to account for that.

src/lib/components/LicenseField.js
alexsdutton: Thank you for the early review, and yep!

src/lib/components/LicenseField.js
alexsdutton: Thanks for the pointers :-).

I'll rename it to RightsField, and have a look at `FundingField` (I had looked at other fields for inspiration, but not that one).

And I'm happy to follow the class-based approach, and am nowhere near well-versed enough in this to make any convincing arguments ;-).

PR #127, opened 2020-12-04
components: add autocompletion for subjects component

audrium:


src/lib/components/LanguagesField.js
```diff
@@ -32,6 +43,7 @@ export class LanguagesField extends Component {
           <FieldLabel htmlFor={fieldPath} icon={labelIcon} label={label} />
         }
         noQueryMessage="Search for languages..."
+        fetchedOptions={fetchedOptions}
```
zzacharo, line 46: Thanks for that :)

audrium:
Here are some examples of how it looks like:

PR #132, opened 2020-12-08
components: fix missing space

ntarocco:


ntarocco:
Warning: not tested PR. I have spotted this during the last November 2020 release demo from @fenekku :)

@zzacharo: I would move the `if isDraftRecord ? <> : null` outside. IMHO it is not responsibility of this component to render empty, unless you are thinking that in the future you will add other stuff to that else.

src/lib/components/FileUploader/FileUploaderToolbar.js
```diff
@@ -63,7 +62,7 @@ export const FileUploaderToolbar = ({
                 ? { color: 'blue' }
                 : {})}
             >
-              {humanReadableBytes(filesSize)} out of
+              {humanReadableBytes(filesSize)} out of{' '}
```
zzacharo, line 65: Maybe to avoid deleting this space we can convert it to `${humanReadableBytes(filesSize)}} out of 
  ${humanReadableBytes(quota.maxStorage)}`

zzacharo (approved):

src/lib/components/FileUploader/FileUploaderToolbar.js
zzacharo: After testing your solution fits better :)

zzacharo:
> Warning: not tested PR. I have spotted this during the last November 2020 release demo from @fenekku :)
> 
> @zzacharo: I would move the `if isDraftRecord ? <> : null` outside. IMHO it is not responsibility of this component to render empty, unless you are thinking that in the future you will add other stuff to that else.

I will include the change for the `isDraftRecord` in my next PR :)

PR #146, opened 2020-12-11
form: refactor form to support partial save

fenekku:
depends on:
- https://github.com/inveniosoftware/invenio-records-resources/pull/186
- https://github.com/inveniosoftware/invenio-rdm-records/pull/321

This is a fair amount of code changes to support partial save and form state.

The mental model to have is that the form is a state machine. It can be saving/publishing, successful after save/publish, partially successful after save and unsuccessful after save/publish.

What do you know, redux is actually well-suited for state machines. So the code should be well suited.

A couple of things are done in passing.

- insulate axios to API client code only.
  This makes the controller uncoupled from the api
  client's internals.
- move error handling from invenio-app-rdm to this
  module. Error handling is coupled to form state so
  makes sense to put it all together.
- provide form feedback if the save is successful

There are many things that will still need work:
- check all cases
- Have API field errors be translated to human readable
  fields.
- deserialize deeply nested errors so that they can
  be assigned to fields via formik correctly.
- Use a top-of-page banner to provide feedback as
  opposed to in-form message.
- ...

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/389

All good save:

form_feedback_saved_good

Partial save:

form_feedback_saved_warning

Publish error:

form_feedback_saved_bad


src/lib/DepositRecordSerializer.js
```diff
@@ -97,6 +98,19 @@ export class DepositRecordSerializer {
     }),
   };
 
+  /**
+   * Return Field matching given fieldpath
+   * @method
+   * @param {string} fieldpath - dotted path to field in metadata
+   * @returns {Field} corresponding field
+   */
+  getField(fieldpath) {
```
zzacharo, line 107: Isn't it the same as `lodash/get` function?

src/lib/actionButtons/PublishButton/PublishButton.js
```diff
@@ -22,12 +22,13 @@ export default class PublishButton extends Component {
...
     const record = formik.values;
-    return (
+    const r = (
       !this.draftAlreadyCreated(record) ||
-      (this.props.filesEnabled && !this.props.nubmerOfFiles) ||
+      (this.props.filesEnabled && !this.props.numberOfFiles) ||
       this.props.fileUploadOngoing ||
       formik.isSubmitting
     );
+    return r;
```
zzacharo, line 31: The return of the assignment is not needed I guess...

src/lib/DepositRecordSerializer.js
fenekku: yes! Must have been the end of the week.

src/lib/actionButtons/PublishButton/PublishButton.js
fenekku: There is a lot of leftover code from debugging. Just a wip, so I wouldn't worry about it.

src/lib/DepositController.js
```diff
@@ -54,47 +57,76 @@ export class DepositController {
...
+      response = await this.createDraft(payload, { store });
+    } else {
+      response = await this.apiClient.save(payload);
+    }
+
+    let data = recordSerializer.deserialize(response.data || {});
+    let errors = recordSerializer.deserializeErrors(response.errors || []);
+
+    // response 100% successful
+    if ( 200 <= response.code && response.code < 300 && _isEmpty(errors) ) {
```
zzacharo, line 71: Maybe extract this condition in the `DepositApiClientResponse` class and then you do something like `if response.is_ok()`. WDYT?

src/lib/DepositController.js
```diff
@@ -54,47 +57,76 @@ export class DepositController {
...
+      store.dispatch({
+        type: ACTION_SAVE_SUCCEEDED,
+        payload: { data },
+      });
+    }
+    // response partially successful
+    else if (200 <= response.code && response.code < 300 ) {
```
zzacharo, line 78: Same here :)

src/lib/DepositFormFeedback.js
```diff
@@ -0,0 +1,105 @@
...
+      // TODO: Use some vocabulary(?) to get human readable label
+      const fieldLabel = k;
+      const value = errors.metadata[k];
+      let fieldErrorMessage;
+      if (_isString(value)) {
+        fieldErrorMessage = value;
+      } else {
+        // WHY: value may be an object with sub-objects, but we want the leaves
+        //      i.e. all error messages for a top-level field in a single
+        //      string (for now).
```
zzacharo, line 70: Maybe a practical example here?

src/lib/DepositFormFeedback.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+
+const mapStateToProps = (state) => ({
+  formAction: state.deposit.formAction,
+  errors: state.deposit.errors,
+});
+
+
+export const DepositFormFeedback = connect(
```
zzacharo, line 102: Maybe rename this to `FormFeedback` and move it under the `components` folder? Also, to keep consistency I would add the connect to an index.js and make this a folder....Otherwise, we end up with an inconsistent way of declaring components that need redux....

src/lib/DepositRecordSerializer.js
zzacharo: Forgot to move to `lodash/get` :)

src/lib/actionButtons/PublishButton/PublishButton.js
zzacharo: Reminder for the redundant return assignment :)

src/lib/components/ResourceTypeField.js
```diff
@@ -37,6 +37,10 @@ export class ResourceTypeField extends Component {
       (e) => e['parent-value'] === resource_type
     );
 
+    // if (subtypeOptions.length < 1) {
```
zzacharo, line 40: Is this a leftover?

src/lib/components/ResourceTypeField.js
```diff
@@ -50,7 +54,7 @@ export class ResourceTypeField extends Component {
             />
           }
           options={this.props.options.type}
-          onChange={handleChange}
+          // onChange={handleChange}
```
zzacharo, line 57: Leftover?

src/lib/index.js
```diff
@@ -8,6 +8,7 @@
 export { DepositApiClient } from './DepositApiClient';
 export { DepositController } from './DepositController';
 export { DepositFormApp } from './DepositFormApp';
+export { DepositFormFeedback } from './DepositFormFeedback';
```
zzacharo, line 11: See my comment for moving this to `components` and maybe rename it to `FormFeedback`

src/lib/state/actions/deposit.js
```diff
@@ -46,23 +34,22 @@ export const save = (record, formik) => {
 
 export const submitAction = (action, event, formik) => {
   return async (dispatch, getState, config) => {
-    console.log(`onSubmit - ${action}`);
     dispatch({
       type: FORM_ACTION_EVENT_EMITTED,
       payload: action,
     });
-    formik.handleSubmit(event);
+    formik.handleSubmit(event); // eventually calls submitFormData below
```
zzacharo, line 41: Useful comment :)

src/lib/state/types/index.js
```diff
@@ -6,9 +6,6 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 export const FORM_ACTION_EVENT_EMITTED = 'FORM_ACTION_EVENT_EMITTED';
```
zzacharo, line 8: Maybe add these below with the other form action types?

zzacharo (commented):
Nice job!!!!

src/lib/DepositRecordSerializer.js
```diff
@@ -190,6 +228,13 @@ export class DepositRecordSerializer {
       );
     }
     // Remove empty values again because serialization may add some back
-    return this.removeEmptyValues(serializedRecord);
+    serializedRecord = this.removeEmptyValues(serializedRecord);
```
zzacharo, line 231: I think would be cleaner in some point to ensure that the serializer doesn't set empty values instead of removing them again....

fenekku (commented):
Thanks for the review! Will merge and release.

PR #153, opened 2021-01-15
452 delete discard button

fenekku:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/452

### DIFF
--- a/src/lib/DepositApiClient.js
+++ b/src/lib/DepositApiClient.js
@@ -37,16 +37,34 @@ export class DepositApiClient {
     this.createUrl = createUrl;
   }
 
+  async createResponse(axios_call) {
+    try {
+      let response = await axios_call();
+      return new DepositApiClientResponse(
+        response.data,  // exclude errors?
+        response.data.errors,
+        response.status
+      );
+    } catch (error) {
+      console.dir("error", error);
+      return new DepositApiClientResponse(
+        error.response.data,
+        error.response.data.errors,
+        error.response.status
+      );
+    }
+  }
+
   /**
    * Calls the API to create a new draft.
    *
-   * @param {object} record - Serialized record
+   * @param {object} draft - Serialized draft
    */
-  async create(record) {
+  async create(draft) {
     try {
       let response = await axios.post(
         this.createUrl,
-        record,
+        draft,
         { headers: { 'Content-Type': 'application/json' } }
       );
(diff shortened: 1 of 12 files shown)


src/lib/state/actions/deposit.js
```diff
@@ -56,3 +56,23 @@ export const submitFormData = (record, formik) => {
...
+ * Returns the async function that controls draft deletion.
+ *
+ * This function is different from the save/publish above because this thunk
+ * is independent of form submission.
+ *
+ * @param {object} event - click event
+ * @param {object} formik - formik object
+ */
+export const discard = (event, formik) => {
+  return async (dispatch, getState, extra) => {
```
zzacharo, line 70: The 3rd parameter was called `config` as it includes configuration specific to the deposit module e.g the controller class...For consistency, I would keep the `config` name...

zzacharo (approved):
You can merge after addressing the comments!

src/lib/DepositApiClient.js
```diff
@@ -37,24 +37,16 @@ export class DepositApiClient {
...
-        { headers: { 'Content-Type': 'application/json' } }
-      );
+      let response = await axios_call();
       return new DepositApiClientResponse(
         response.data,  // exclude errors?
         response.data.errors,
         response.status
       );
     } catch (error) {
+      console.dir("error", error);
```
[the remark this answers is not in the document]
fenekku: yep! Thanks!

src/lib/state/actions/deposit.js
fenekku: This code is what makes it very confusing: https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/store.js#L52 : we pass an `appConfig` to the store that holds a `config` variable and an `apiConfig` except the `apiConfig` is not just an API configuration it is also a controller and a serializer and the `config` variable (accessible by `getState().deposit.config`) well... sounds a lot like the deposit config but is different :sweat_smile: .  

What I want to do in a next PR is to replace the `apiConfig` by `controller` (controller instance) and have it hold the serializer and api client.

To be fair, I don't know of a "good approach" to divide up our code in react/redux (should we be using a controller like we are in the way we are?). I think we may have a better perspective over time when some of the semantic is cleaned up. There must also exist good examples out there (at the end of the day this is typical JS app stuff we are doing right?).

Issue #147, opened 2021-01-05
ux: mark as metadata-only clears form if not saved


fenekku:
Package version (if known): december release

## Describe the bug
A completed form can be cleared out completely if file upload component is not dealt with first. At first blush, this is probably due to file upload component creating a draft without passing metadata and then the form's values being filled automatically by the returned metadata from the API (i.e. empty object).

## Steps to Reproduce
1. Go to '/uploads/new'
2. Fill out the metadata portion of the form (don't touch the file upload portion)
3. Realize that you need to mark the record as metadata-only, so click on the metadata-only check
4. See the whole form being cleared.

## Expected behavior
Filling the form should be a free-flowing experience where filling different parts at different times / in different order shouldn't matter. 

## Additional context
This is bug I was mentioning yesterday @lnielsen .

PR #156, opened 2021-01-20
files: don't clear form when clicking metadata-only

fenekku:
- closes https://github.com/inveniosoftware/react-invenio-deposit/issues/147
- hides rest of file upload when metadata-only is clicked

image
(font size change in invenio-app-rdm)

- refactors some code a little / more to come

### DIFF
--- a/src/lib/DepositApiClient.js
+++ b/src/lib/DepositApiClient.js
@@ -118,6 +118,8 @@ export class DepositApiClient {
     );
   }
 
+  // TODO: Might consider extracting these out to a FilesApiClient.js
+
   initializeFileUpload(initializeUploadUrl, filename) {
     const payload = [
       {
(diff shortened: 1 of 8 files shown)


src/lib/DepositApiClient.js
```diff
@@ -118,6 +118,8 @@ export class DepositApiClient {
     );
   }
 
+  // TODO: Might consider extracting these out to a FilesApiClient.js
```
zzacharo, line 121: comment: I didn't do that as it was just one function and also the depositApiClient is configurable so in case you want to override the class you just have everything there...maybe instead of TODO create an issue? Easier to track them :)

zzacharo (approved):

src/lib/DepositApiClient.js
fenekku: :+1: I've added an issue here: https://github.com/inveniosoftware/react-invenio-deposit/issues/157 . This was a case of a little piece of thread that unraveled much more once I pulled on it.

PR #158, opened 2021-01-21
resource_type field: merge dual dropdown into one [+]

### LINKED ISSUE #78: Convert the dual resource type dropdowns into a single dropdown field
If I recall correctly, @lnielsen wanted to collapse the 2 dropdowns into one. This will make that field take less space and allow us to reuse it elsewhere in the form nicely (see Related Work field).

## Describe the solution you'd like

A single dropdown with resource type choices in the form of

`<General> / <Subtype>`

fenekku:
- closes #78 

Also include a lot of improvements:

- Use FastField (quite satisfying to see this field's re-rendering
  NOT triggered by every keystroke in another field).
- Display icon of selected option in the selected field
- Display error correctly

**Screenshots**

Icons:
resourcetype_dropdown

Error:
resourcetype_error

### DIFF
--- a/src/lib/components/ResourceTypeField.js
+++ b/src/lib/components/ResourceTypeField.js
@@ -1,81 +1,114 @@
 // This file is part of React-Invenio-Deposit
-// Copyright (C) 2020 CERN.
-// Copyright (C) 2020 Northwestern University.
+// Copyright (C) 2020-2021 CERN.
+// Copyright (C) 2020-2021 Northwestern University.
 //
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Field } from 'formik';
+import { FastField } from 'formik';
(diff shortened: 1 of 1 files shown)


src/lib/components/ResourceTypeField.js
```diff
@@ -1,81 +1,114 @@
...
+            values: {type: type.value, subtype: subtype.value}
+          }))
+        );
+      } else {
+        // Push an option corresponding to the type only
+        options.push({
+            icon: type.icon,
+            text: type.text,
+            value: type.value,
+            values: {type: type.value}
```
zzacharo, line 49: Just out of curiosity...the dropdown options can also include arbitrary extra data?

zzacharo (approved):

src/lib/components/ResourceTypeField.js
fenekku: Yep, they are just objects. When the user clicks on the option, the `data` object that the click handler receives has `value` only set to the `value` of the option, but we can get back the associated other data by using that `value` as a unique key in the `options` array.

PR #179, opened 2021-01-27
files: upload current form if unsaved when uploading a file

### LINKED ISSUE #178: uploading a file after filling form but not saving clears the form
This is the same bug as the already solved [metadata-only situation](https://github.com/inveniosoftware/react-invenio-deposit/issues/147) but for file upload.

Package version (if known): december release

**Describe the bug**

A completed form can be cleared out completely if a file upload is done before the form is saved.

**Steps to Reproduce**

    Go to '/uploads/new'
    Fill out the metadata portion of the form (don't touch the file upload portion)
    Realize that you want to add a file, so you upload a file
    See the whole form being cleared.

**Expected behavior**

Filling the form should be a free-flowing experience where filling different parts at different times / in different order shouldn't matter.

fenekku:
closes #178

### DIFF
--- a/src/lib/components/FileUploader/FileUploader.js
+++ b/src/lib/components/FileUploader/FileUploader.js
@@ -4,7 +4,8 @@
 //
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
-import React, { Component } from 'react';
+import { useFormikContext } from 'formik';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { Grid, Icon, Message } from 'semantic-ui-react';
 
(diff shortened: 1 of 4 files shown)


src/lib/components/FileUploader/FileUploader.js
```diff
@@ -4,125 +4,132 @@
...
+  filesEnabled,
+  isDraftRecord,
+  quota,
+  record,
+  setFilesEnabled,
+  uploadFilesToDraft,
+  ...uiProps
+}) => {
+  // We extract the working copy of the draft stored as `values` in formik
+  const { values: formikDraft } = useFormikContext();
```
zzacharo, line 32: Just an observation: We have now a mixed way of getting the values from formik i.e hook vs class approach...We should try to converge to one if possible...

zzacharo (approved):

src/lib/components/FileUploader/FileUploader.js
fenekku: yeah... the difference I see is that in some cases we have actual fields that will need to display errors and are "real" inputs to the metadata form and in other cases we have fields that need access to the metadata form but are not inputs to the form (don't need to have other field properties). Agreed that thinking about this and establishing good practices will help.

PR #196, opened 2021-02-04
deposit: helptexts for fields

mb-wali:
closes inveniosoftware/invenio-app-rdm#453

*screenshots*
publication-date

version

relatedWorks

### DIFF
--- a/src/lib/components/RelatedWorksField.js
+++ b/src/lib/components/RelatedWorksField.js
@@ -1,6 +1,7 @@
 // This file is part of React-Invenio-Deposit
 // Copyright (C) 2020-2021 CERN.
 // Copyright (C) 2020-2021 Northwestern University.
+// Copyright (C) 2021 Graz University of Technology.
 //
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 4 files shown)


src/lib/components/RelatedWorksField.js
```diff
@@ -26,6 +27,10 @@ export class RelatedWorksField extends Component {
 
     return (
       <>
+        <label className="helptext">
+          Specify identifiers of related works. Supported identifiers include DOI, Handle, ARK, \
+          PURL, ISSN, ISBN, PubMed ID, PubMed Central ID, ADS Bibliographic Code, arXiv, Life Science Identifiers (LSID), EAN-13, ISTC, URNs, and URLs.'
+        </label>
```
fenekku, line 33: Looking at the screenshot it seems like some bottom padding on this one would help separate it from the ArrayField, no?

fenekku (commented):
The Dates field had a helptext before but it's spacing is not quite the same as those ones. We should align all helptext spacing to be consistent. My preference is to reduce the space between an input and the helptext below it (and increase the space when the help text is not just for the field, but for the section).

mb-wali:
> The Dates field had a helptext before but it's spacing is not quite the same as those ones. We should align all helptext spacing to be consistent. My preference is to reduce the space between an input and the helptext below it (and increase the space when the help text is not just for the field, but for the section).

Strange because i did not touch the css of that see [here](https://github.com/inveniosoftware/invenio-app-rdm/pull/502) - only the target.  i will double check.

zzacharo:
@mb-wali my only concern is regarding the "Related works" helptext...I would put some bottom space between the helptext and the field below...

mb-wali:
> @mb-wali my only concern is regarding the "Related works" helptext...I would put some bottom space between the helptext and the field below...

see here - https://github.com/inveniosoftware/invenio-app-rdm/pull/545#issue-568159871 , more than this?

mb-wali:
> @mb-wali my only concern is regarding the "Related works" helptext...I would put some bottom space between the helptext and the field below...

ah i see it, it did not work for related work, i will try to add inline css in react.

mb-wali:
better? 
related-spacing

PR #197, opened 2021-02-05
resoucetypefield: create label without FieldLabel

### LINKED ISSUE #88: ux: bold/visually differentiate the top-level form fields
The top-level form fields have the same font characteristics as the sub fields. This makes it especially hard to differentiate between top-level required fields and sub-level required fields (i.e. fields that are required only if filled).

We should make top-level fields visually different somehow.

mb-wali:
(note) This component is used in both (top-level) and (sub-level) fields.
closes #88

### DIFF
--- a/src/lib/components/ResourceTypeField.js
+++ b/src/lib/components/ResourceTypeField.js
@@ -72,6 +72,7 @@ export class ResourceTypeField extends Component {
       labelIcon,
       options,
       required,
+      labelClassName,
       ...uiProps
     } = this.props;
 
@@ -107,6 +108,7 @@ export class ResourceTypeField extends Component {
             htmlFor={fieldPath}
             icon={labelIcon}
             label={label}
+            className={labelClassName}
           />
         }
         name={fieldPath}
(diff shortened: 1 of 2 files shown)


src/lib/components/ResourceTypeField.js
```diff
@@ -103,11 +103,10 @@ export class ResourceTypeField extends Component {
...
         label={
-          <FieldLabel
-            htmlFor={fieldPath}
-            icon={labelIcon}
-            label={label}
-          />
+          <label htmlFor={fieldPath}>
+          {labelIcon ? <Icon name={labelIcon} /> : null}
+          {label}
+        </label>
```
fenekku, line 109: The problem is that this "exception" will come up again for Languages and Identifiers and who knows what other fields we may get in the future. Perhaps a more encompassing strategy:

1- See if there is a css rule pattern that can target labels of subfields (`.fielfd > .field label` or something like it) to override those.
2- If there isn't a nice pattern like this across the fields, then make the react-invenio-forms `FieldLabel` take a prop to specify if it's a top-level field or not. The default being the most common current case, so that we don't have to change every instance (Like you were saying, I think that default is it being used for top-level fields.)
?

If that's too vague, let's have a chat about it :smile:

mb-wali:
closing the PR - these changes are not needed.

zzacharo (approved):

PR #219, opened 2021-02-16
218 protection widget connect redux

### LINKED ISSUE #218: form: access: connect to redux for files
The protection widget is connected to formik which makes it connected to the backend. It isn't connected to the presence or absence of files yet. Do so, in order for frontend visuals to be complete.

fenekku:
- closes #218 
- still need to do full manual testing of flow to confirm everything after

### DIFF
--- a/src/lib/components/Access/utils.js
+++ b/src/lib/components/Access/utils.js
@@ -4,87 +4,17 @@
 //
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
-import React, { Component } from 'react';
+import React from 'react';
 import { DateTime } from 'luxon';
-import { Button, Checkbox, Divider, Icon, List, Message } from 'semantic-ui-react';
+import { Divider, Icon, List, Message } from 'semantic-ui-react';
 import {
-  FieldLabel,
   TextAreaField,
-  TextField,
 } from 'react-invenio-forms';
 
-import { Embargo, EmbargoState } from './Embargo';
-import { FastField, Field } from 'formik';
-
-
-class ProtectionButtonsComponent extends Component {
-  /**
-   * Returns the props for a protection button.
-   * @param active is button active
-   * @param activeColor button color when active
-   */
-  getButtonProps(active, activeColor) {
-    let props = {active};
-    if (active) {
-      props["color"] = activeColor;
-    }
-    return props;
-  }
-
-  render() {
-    const {
-      formik,
-      active,
-    } = this.props;
-
-    return (
-      <Button.Group widths={'2'}>
-        <Button
-          {...this.getButtonProps(active, "green")}
-          onClick={(event, data) => {
-            formik.form.setFieldValue(formik.field.name, "public");
-            // NOTE: We reset values, so if embargo filled and click Public,
-            //       user needs to fill embargo again. Otherwise lots of
-            //       bookkeeping.
-            formik.form.setFieldValue("access.embargo", {
-              active: false
-            });
-          }}
-          compact
-          attached
-        >
-          Public
-        </Button>
-        <Button
-          {...this.getButtonProps(!active, "red")}
-          onClick={(event, data) => formik.form.setFieldValue(formik.field.name, "restricted")}
-          compact
-          attached
-        >
-          Restricted
-        </Button>
-
-      </Button.Group>
-    );
-  }
-}
-
-
-export class ProtectionButtons extends Component {
-  render() {
-    const {
-      fieldPath,
-      ...props
-    } = this.props;
-
-    return <FastField
-      name={fieldPath}
-      component={(formikProps) => (
-        <ProtectionButtonsComponent formik={formikProps} {...props} />
-      )}
-    />;
-  }
-}
+import { EmbargoState } from './Embargo';
+import { ProtectionButtons } from './ProtectionButtons';
+import { EmbargoCheckboxField } from './EmbargoCheckboxField';
+import { EmbargoDateField } from './EmbargoDateField';
 
 
 export function MetadataSection({isPublic}) {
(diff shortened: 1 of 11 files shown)


src/lib/components/Access/utils.js
```diff
@@ -13,44 +13,93 @@ import {
...
+            });
+          }}
+          compact
+          attached
+        >
+          Public
+        </Button>
+        <Button
+          {...this.getButtonProps(!active, "red")}
+          onClick={(event, data) => formik.form.setFieldValue(formik.field.name, "restricted")}
```
zzacharo, line 60: minor: Just to follow the same patterns as in the rest of the codebase maybe we can use here the passed prop `fieldPath`.

src/lib/components/Access/utils.js
```diff
@@ -78,6 +127,57 @@ export function MessageSection({intent, icon, title, text}) {
...
+    return <Checkbox
+      disabled={embargo.is(EmbargoState.DISABLED)}
+      checked={embargo.is(EmbargoState.APPLIED)}
+      onChange={
+        (event, data) => {
+          if (formik.field.value) {
+            // NOTE: We reset values, so if embargo filled and user unchecks,
+            //       user needs to fill embargo again. Otherwise lots of
+            //       bookkeeping.
+            formik.form.setFieldValue("access.embargo", {
```
zzacharo, line 147: Why not follow the same pattern as below? i.e `formik.form.setFieldValue("access.embargo.active", false);`

src/lib/components/Access/utils.js
```diff
@@ -104,8 +204,7 @@ export function embargoSection(embargo) {
     opacity: "0.5",
     cursor: "default !important"
   }
-  const disabled = embargo.is(Embargo.DISABLED) || embargo.is(Embargo.LIFTED);
-  const embargoStyle = disabled ? greyedStyle : {};
+  const embargoStyle = embargo.is(EmbargoState.DISABLED) ? greyedStyle : {};
```
zzacharo, line 207: about the style: I think if we start putting also inline styles we will mess more the UX general styling and we will end up not knowing from where the css rules are coming from....I guess also for consistency we can define for lists are own "greyStyle" i.e disabled view I guess and use it widely...

src/lib/components/AccessRightField.js
```diff
@@ -49,37 +47,49 @@ class Protection {
 }
 
 
-export class AccessRightField extends Component {
+function convertToNewFormat(access) {
+  let record;
+  if (access.record && ["public", "restricted"].includes(access.record)) {
```
zzacharo, line 52: That first check is redundant as worst case the `access.record` will returned undefined which satisfies the second check I think...

src/lib/components/AccessRightField.js
```diff
@@ -49,37 +47,49 @@ class Protection {
...
+      fieldPath,
+      formik,  // this is our access to the shared current draft
+      hasFiles,
+      label,
+      labelIcon,
+    } = this.props;
+
+    // value of access field
+    // temporarily convert to upcoming backend format
+    const access = convertToNewFormat(formik.field.value);
```
zzacharo, line 91: I see the pattern you use here with the `formik.field.value` and I am wondering if we should use it in the other components to avoid `getIn(formik.form.values, fieldPath, defaultValue)` or stick to the old way for consistency..... My main concern is that we are starting to create different habits of doing things and codebase is getting messy.... Maybe we create an issue to evaluate that?

src/lib/components/AccessRightField.js
```diff
@@ -111,13 +121,47 @@ export class AccessRightField extends Component {
...
 
-AccessRightField.defaultProps = {
-  fieldPath: 'access.access_right',
+FormikAccessRightField.defaultProps = {
+  fieldPath: 'access',
 };
+
+
+const mapStateToProps = (state) => ({
+  hasFiles: Object.values(state.files.entries).length > 0,
```
zzacharo, line 160: I think that now we have more places that this transformation occurs so maybe we should just store it in the redux state once a file action is triggered instead...

src/lib/components/AccessRightField.js
```diff
@@ -111,13 +121,47 @@ export class AccessRightField extends Component {
   }
 }
 
-AccessRightField.propTypes = {
+class FormikAccessRightField extends Component {
```
zzacharo, line 124: This can be renamed to just `AccessRrightField` same fashion as the other components...

src/lib/components/AccessRightField.js
```diff
@@ -111,13 +121,47 @@ export class AccessRightField extends Component {
...
+  render() {
+    // NOTE: This is a "cute" optimization.
+    //       In general, FastField only re-renders if
+    //       * formik slice associated with this.props.fieldPath changes
+    //         (i.e. `access` changes)
+    //       * props are ADDED or REMOVED to FastField
+    // So we add/remove a prop to FastField based on the presence of files.
+    // This way, FastField only renders when the things (access and hasFiles)
+    // it cares about change as it should be.
+    const change = this.props.hasFiles ? {change: true} : {}
```
zzacharo, line 134: Comment: Out of curiosity, the `hasFiles` prop is coming from redux so I would guess that the `FormikAccessRightFIeld` component is being rerendered when the prop is changing and thus I would expect the ....Have you checked that the FastField is not rerendered without passing this custom prop? Actually by checking the code of [FastField](https://github.com/formium/formik/blob/master/packages/formik/src/FastField.tsx#L75) I don't see how that could work unless you pass the custom `shouldUpdate` prop and change the way your component is being updated...

src/lib/components/Access/utils.js
fenekku: Sure. I do find relying on `formik.field.name` to be slightly cleaner overall in that it reduces the number of passed props.

src/lib/components/Access/utils.js
fenekku: ... because of the comment above :smile_cat: ... but maybe the comment is not clear enough:

In this situation we need to wipe away the `until` and `reason` keys since the `Apply embargo` is not selected anymore. We do that by setting the whole object as opposed to a single field in the object as is done below. Otherwise we pass inconsistent embargo values under the hood and the backend will return an error that doesn't match up to what the user was expecting.  

In the case below, selecting the "Apply embargo" has no other responsibilities (it doesn't need to reset anything), so it just sets it directly.

src/lib/components/Access/utils.js
fenekku: Would it be better if I move this style to invenio-app-rdm deposit.less and here I will set the class corresponding to it?

src/lib/components/AccessRightField.js
fenekku: yes! And now that the backend is fine, I should be able to remove this alltogether.

src/lib/components/Access/utils.js
zzacharo: I would say that this can be part of the `list.overrides` of the `rdm` theme as it is indeed a generic feature for the list components....On the other hand, it might be used only in this list so putting it in the `deposit.less` might also make sense...

src/lib/components/Access/utils.js
zzacharo: Ok I see!

src/lib/components/Access/utils.js
zzacharo: I am ok with it if we normalize it everywhere even I am afraid that might not be feasible everywhere.....I think is worth logging an issue to reiterate through the components and see if this can be applied, especially if it cleans up the props needed!

src/lib/components/AccessRightField.js
fenekku: :grin: Ha! This is going away, so it should be fine. The other place using `formik.field.value` is in utils and it's for the reason you mentioned: saves us the trouble of trying to find the field value ourselves when Formik already gives it to us.

About the consistency, I think I agree with tending towards consistency long-term and being self-consistent within a PR. But I don't think I agree with consistency short term. In order to improve things, we need to be able to "run experiments" and do things differently if we think it's better. However it's hard to know if some ways are better right away, so retrofitting immediately in the name of consistency lends itself to "rehacktoring" (churn in place). 

Unfortunately, I don't know how to operationalize this: my experience has been that "refactoring" tickets don't work (nobody takes them because they are not "needed" - 'hey it works, don't touch it'). The boy scout rules seems to work better, but when do we know/communicate that a pattern is good enough to be spread around? Seems like an organic process in the end. For example, I think the pattern you've introduced for Formik wrapper:

```js
      <Field
        name={this.props.fieldPath}
        component={(formikProps) => ( 
          <WrappedComponent formik={formikProps} {...this.props} />
        )}
      />
```
is really nice because of readability and clear separation and I am re-using it now.

src/lib/components/AccessRightField.js
fenekku: We can't here because of the double-wrapping:

Raw component: `AccessRightFieldComponent ->` Formik wrap: `FormikAccessRightField->` Redux wrap and Final name: `AccessRightField`

src/lib/components/AccessRightField.js
fenekku: I had checked. It's the React shouldComponentUpdate rules in action: https://reactjs.org/docs/optimizing-performance.html#shouldcomponentupdate-in-action  . The gist of it is: when props/state changes, React always re-assesses if the component should be re-rendered, BUT tt doesn't always re-render the component; it's assessment might be: "no need to re-render". And FastField makes it so, unless for these reasons:  https://formik.org/docs/api/fastfield#overview

zzacharo (approved):

PR #223, opened 2021-02-22
FormFeedback: hide ul if no errors

fenekku:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/393
- removes the Formik wrapper around Feedback since unnecessary (redux provides everything): helps reduce lag in form interaction.

### DIFF
--- a/src/lib/components/FormFeedback.js
+++ b/src/lib/components/FormFeedback.js
@@ -1,11 +1,10 @@
 // This file is part of InvenioRDM
-// Copyright (C) 2020 CERN.
-// Copyright (C) 2020 Northwestern University.
+// Copyright (C) 2020-2021 CERN.
+// Copyright (C) 2020-2021 Northwestern University.
 //
 // Invenio App RDM is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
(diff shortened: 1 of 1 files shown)


src/lib/components/FormFeedback.js
```diff
@@ -94,17 +93,12 @@ class DisconnectedFormFeedback extends Component {
         <Grid container>
           <Grid.Column width={15} textAlign="left">
             <Message.Header>{message}</Message.Header>
-            <Message.List>{listErrors}</Message.List>
+            { Boolean(listErrors.length) && <Message.List>{listErrors}</Message.List> }
```
zzacharo, line 96: `listErrors.length && ...` should be enough, no?

fenekku: I thought so too at first! But I learned that React only doesn't render `false` values. So in the case of `listErrors.length && ...`, React renders `0` if the list is empty. Therefore the need for the `Boolean`.

fenekku: I pushed `listErrors.length > 0` to be less weird :)

zzacharo: Interesting didn't know that! Thanks for the explanation :)

PR #233, opened 2021-02-25
versionField: adds link to helptext

mb-wali:
closes inveniosoftware/invenio-app-rdm#590
version


fenekku (approved):
was it the `span` that did it?

PR #242, opened 2021-03-22
deposit: refine form title/delete button text

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/704

Screenshot 2021-03-22 at 15 40 20
Screenshot 2021-03-22 at 15 40 41
Screenshot 2021-03-22 at 15 40 57

### DIFF
--- a/src/lib/components/DeleteButton.js
+++ b/src/lib/components/DeleteButton.js
@@ -13,6 +13,24 @@ import { ActionButton } from 'react-invenio-forms';
 import { discard } from '../state/actions';
 import { toCapitalCase } from '../utils';
 
+const DialogText = ({ action }) => {
+  let text = '';
+  switch (action) {
+    case 'discard changes':
+      text = 'Are you sure you want to discard the changes to this draft?';
+      break;
+    case 'discard version':
+      text = 'Are you sure you want to delete this new version?';
+      break;
+    case 'delete':
+      text = 'Are you sure you want to delete this draft?';
+      break;
+    default:
+      break;
+  }
+  return text;
+};
+
 export class DeleteButtonComponent extends Component {
   state = { modalOpen: false };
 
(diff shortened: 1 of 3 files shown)


src/lib/components/DeleteButton.js
```diff
@@ -80,6 +110,8 @@ export class DeleteButtonComponent extends Component {
 
 const mapStateToProps = (state) => ({
   isSaved: Boolean(state.deposit.record.id),
+  isPublished: state.deposit.record.is_published,
+  isNewVersion: state.deposit.record.versions?.index > 1,
```
fenekku, line 114: I got tripped up by this and maybe others might be too :sweat_smile: : `isNewVersion` -> `isVersion` could help make it clear that this value doesn't mean it's necessarily a new version, but rather that it is a (subsequent) version of a record (new only if it hasn't been published)

src/lib/components/DepositFormTitle.js
```diff
@@ -5,14 +5,27 @@
...
+    } else {
+      content = 'Edit upload';
+    }
+    return <Header as="h1" icon="upload" content={content} />;
   }
 }
+
+const mapStateToProps = (state) => ({
+  isPublished: state.deposit.record.is_published,
+  isNewVersion: state.deposit.record.versions?.index > 1,
```
fenekku, line 25: Same here. Also `versions.index` is guaranteed no? (we could do without `?.`)

fenekku (commented):
Minor wording and a rename to make it clearer :+1:

src/lib/components/DepositFormTitle.js
zzacharo: No, because in case of a new draft (never saved it before) we get only empty values :) I could check in other ways if version exists, but I find this feature more elegant for this 😅  But, I had in mind what we discussed in the past and I am making sure I am using it only if the value is not guaranteed!

fenekku: Good to know, thanks!

Issue #250, opened 2021-03-31
Avoid creation of multiple drafts


ntarocco:
Package version (if known): 0.13.2

## Describe the bug
After clicking on `New upload`, the form appears. If you click on the checkbox `Metadata-only record`, the web page will take some time to create a new draft and "redirect" there. Since I tested it with a slower connection, after clicking nothing was happening and I clicked it again.
As result, 2 drafts were created.

## Steps to Reproduce
1. Go to 'Uploads' page
2. Click on 'New upload'
3. Slow down your network connection
4. Click quickly 2 times on the checkbox "Metadata-only record" and observe the URL.
5. Come back to 'Uploads' and you should see 2 uploads.

## Expected behavior

After any action that will trigger the creation of the first draft (from `/uploads/new` to `/uploads/zc2t5-1a468`, while creating the new draft the form should be disabled in a loading state (showing a loading) to avoid double submission.

## Screenshots (if applicable)

![Screenshot 2021-03-31 at 15 50 55](https://user-images.githubusercontent.com/2089455/113156217-13a0dd00-923a-11eb-832d-30f6ced93a94.png)

# EDIT 2021-04-06
The problem is more widespread and reproducible without throttling. 

## Steps to Reproduce
1. Go to 'Uploads' page
2. Click on 'New upload'
3. Double-click on "Save Draft" button
4. Come back to 'Uploads' and you should see 2 uploads.

Investigation so far:

We use `isSubmitting` to disable the `SaveButton`, but although it is briefly `true`, it **somehow** reverts to `false` and lets the second click go through. Spent all day to discover this "somehow" (validation could have been a culprit but `isValid` is always `true`) to no avail. At end of day found these bug reports: 
- https://github.com/formium/formik/issues/1730
- https://github.com/formium/formik/issues/2442  

but I thought `dispatch` call was synchronous after talking with Nico. ... :confounded:

fenekku:
Actually thanks for that! I think it's an argument in favour of investigating if "Metadata-only check" could be passed as a field in the draft JSON rather than as a separate API call. If we manage that, we can solve that problem, improve the UX and reduce the load on our servers. Exciting!

ntarocco:
> Actually thanks for that! I think it's an argument in favour of investigating if "Metadata-only check" could be passed as a field in the draft JSON rather than as a separate API call. If we manage that, we can solve that problem, improve the UX and reduce the load on our servers. Exciting!

[...]

PR #253, opened 2021-04-01
deposit-form: show warning to user if files are missing while saving draft

audrium:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/736
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/456
Provides some feedback to the end-user that files are missing while trying to save the draft. Result:
pr3


fenekku (approved):
Nice! In light of https://github.com/inveniosoftware/react-invenio-deposit/issues/250 we may want to revisit the problem area down the road. This is simple and unobstrusive.

PR #258, opened 2021-04-08
buttons: Fixes loading state

jrcastro2:
Fix loading state of the buttons

loading-buttons

### DIFF
--- a/src/lib/components/DeleteButton.js
+++ b/src/lib/components/DeleteButton.js
@@ -31,14 +31,15 @@ const DialogText = ({ action }) => {
 };
 
 export class DeleteButtonComponent extends Component {
-  state = { modalOpen: false };
+  state = { modalOpen: false, isDeleting: false };
 
   handleOpen = () => this.setState({ modalOpen: true });
 
   handleClose = () => this.setState({ modalOpen: false });
 
   isDisabled = (formik) => {
-    return !this.props.isSaved || formik.isSubmitting;
+    const { isDeleting } = this.state;
+    return !this.props.isSaved || formik.isSubmitting || isDeleting;
   };
 
   render() {
@@ -49,10 +50,11 @@ export class DeleteButtonComponent extends Component {
       isVersion,
       ...uiProps // only has ActionButton props
     } = this.props;
-
+    const { isDeleting } = this.state;
     const handleDelete = (event, formik) => {
       deleteClick(event, formik);
       this.handleClose();
+      this.setState({ isDeleting: true });
     };
 
     let action = '';
@@ -77,7 +79,11 @@ export class DeleteButtonComponent extends Component {
         >
           {(formik) => (
             <>
-              <Icon name="trash alternate outline" />
+              {formik.isSubmitting && isDeleting ? (
+                <Icon size="large" loading name="spinner" />
+              ) : (
+                <Icon name="trash alternate outline" />
+              )}
               {capitalizedAction}
             </>
           )}
--- a/src/lib/components/PreviewButton.js
+++ b/src/lib/components/PreviewButton.js
@@ -24,6 +24,10 @@ export class PreviewButtonComponent extends Component {
     previousFormState: '',
   };
 
+  isDisabled = (formik) => {
+    return formik.isSubmitting;
+  };
+
   render() {
     const { record, saveClick, formState, ...uiProps } = this.props;
     const { isLoading, previewButtonClicked, previousFormState } = this.state;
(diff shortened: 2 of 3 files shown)


src/lib/components/DeleteButton.js
```diff
@@ -50,10 +51,11 @@ export class DeleteButtonComponent extends Component {
       isVersion,
       ...uiProps // only has ActionButton props
     } = this.props;
-
+    const { isDeleting } = this.state;
     const handleDelete = (event, formik) => {
       deleteClick(event, formik);
       this.handleClose();
+      this.setState({ isDeleting: true });
     };
```
fenekku, line 59: :detective: : This actually reveals a pre-exisitng problem: we are not dealing with errors in deletion correctly. Clearly here `isDeleting` is set to true forever if the delete didn't work, but the problem is larger: `formik` is not a `formik` object at all! It's a `data: object` containing all the props of `Button` ([reference](https://react.semantic-ui.com/elements/button/#types-button)). `Button` [here](https://github.com/inveniosoftware/react-invenio-deposit/pull/258/files#diff-e76813af199e88111670c54019b3d3c5f7f732f8e8f7329d6ad0d069c133d7c3R107) should be an `ActionButton` to pass a formikBag. It needs more investigation to even see if we need formikBag in this case at all. 

I made a separate issue about all this: https://github.com/inveniosoftware/react-invenio-deposit/issues/259 .

Thanks for triggering this investigation.

src/lib/components/PreviewButton.js
```diff
@@ -63,7 +69,7 @@ export class PreviewButtonComponent extends Component {
       >
         {(formik) => (
           <>
-            {isLoading ? (
+            { ( formik.isSubmitting || isLoading ) ? (
```
fenekku, line 72: up for discussion: Interesting, so if a user clicks on "Save Draft", all other buttons (here we see the pattern on the PreviewButton) won't just be disabled, they will also be spinning? I would just make the clicked button spinning and the rest disabled. The feedback comes out quite strong otherwise and less specific.

src/lib/components/PublishButton.js
```diff
@@ -63,7 +61,7 @@ export class PublishButtonComponent extends Component {
         >
           {(formik) => (
             <>
-              { ( formik.isSubmitting && formState === FORM_PUBLISHING ) ? (
```
fenekku, line 66: See comment above: this was made so that *only* this button is spinning when the form is publishing.

src/lib/components/SaveButton.js
```diff
@@ -33,7 +33,7 @@ export class SaveButtonComponent extends Component {
       >
         {(formik) => (
           <>
-            { ( formik.isSubmitting && formState === FORM_SAVING ) ? (
+            { ( formik.isSubmitting || formState === FORM_SAVING ) ? (
```
fenekku, line 36: same

fenekku (commented):
Thanks for triggering my investigation into the delete UX problems (nothing for you to do on that front right now, unless you want to change Button for ActionButton).

Let's change the spinning to be specific to the button that was clicked and after that we should be good.

src/lib/components/DeleteButton.js
jrcastro2: I see, I added the `isDeleting` and set it true forever because I saw this comment [here](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositController.js#L180-L182) and assumed all good for the moment.

src/lib/components/PreviewButton.js
jrcastro2: Hmm, currently if you click on the "Preview" button the "Save Draft" button will also be spinning and disabled (when we click the "Preview" button we also trigger the save) but the rest don't are disabled nor spin because the `formik.isSubmitting` is set to `False` after the "onSubmit()" (check [here](https://formik.org/docs/guides/form-submission)) but the redux state is still loading. This causes that when we click a button, the `formik.isSubmitting` changes to true for a very short period of time while it stays in the loading state for much more time (until we get the response). I agree with making the clicked buttons spin and have the others disabled but we would need to check for all the redux loading states to disable them since the `formik.isSubmitting` is set to false before redux state is updated. Would this check be OK? or any better ideas?

fenekku: > This causes that when we click a button, the formik.isSubmitting changes to true for a very short period of time

I think this should be solved by https://github.com/inveniosoftware/react-invenio-deposit/pull/257 now. There was indeed a bug where the `formik.isSubmitting` was very short, but now it should be as long as it takes to get the response. So other buttons should be disabled correctly now. Making `formik.isSubmitting && isLoading` the condition for spinning and `formik.isSubmitting` the condition for disabled here should be good, right?

jrcastro2: You are right, `formik.isSubmitting` bug is solved. Your proposal works flawless. Thanks a lot!

fenekku (approved):
:100:

PR #262, opened 2021-04-11
form: implement pids field

### LINKED ISSUE #260: Add PIDsField
Create a new field to handle the `pids` field of the record.
For a given `pid`, it should handle managed and unmanaged PID.
For the DOI, enable DOI reservation.

ntarocco:
* implements new PID field to handle managed and unmanaged PIDs
* closes #260

Screenshot 2021-04-11 at 17 35 23


src/lib/components/AffiliationsField.js
```diff
@@ -5,13 +5,12 @@
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
-import React, { Component } from 'react';
+import _get from 'lodash/get';
```
ppanero, line 8: comment: I don't really mind because I guess this is a correct ordering? And I see it actually compressed many that were in different lines so 💪 . But it seems to be autosave changes? to be careful in code not to alter the git history.

src/lib/components/Identifiers/IdentifiersField.js
```diff
@@ -35,26 +34,24 @@ export class IdentifiersField extends Component {
         >
           {({ array, arrayHelpers, indexPath, key }) => (
             <GroupField widths="equal">
-              {schemeOptions &&
+              {schemeOptions && (
```
ppanero, line 37: question: mostly for my learning, does the parenthesis imply any change on behaviour?

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,466 @@
...
+const PROVIDER_EXTERNAL = 'external';
+const UPDATE_PID_DEBOUNCE_MS = 200;
+
+/**
+ * Button component to reserve a PID.
+ */
+class BtnReservePID extends Component {
+  handleClickReservePID = () => {
+    const { onPIDReserved } = this.props;
+    console.log('TODO implement reserve PID');
```
ppanero, line 25: minor: shall we change by `PIDS-FIXME` like in rdm-records so we can grep and get them all?

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,466 @@
...
+  disabled: false,
+};
+
+/**
+ * Button component to unreserve a PID.
+ */
+class BtnUnreservePID extends Component {
+  unreserve = (identifier) => {
+    const { onPIDUnreserved } = this.props;
+    console.log('TODO implement unreserve PID', identifier);
```
ppanero, line 60: minor: `PIDS-FIXME`

ppanero (approved):
LGTM. A few minors, but mostly styling or doubts on my side to learn :) In terms of behaviour, to clarify (and update the RFC):

- Managed: There is no way for the user to set up client/provider, because those are set up by the system.
- Unmanaged: Provider is always `external` and we do not care about the `client`.
- Accepted PIDs type (dict keys at datamodel level) are always defined by the system (i.e. amount of fields in the deposit form). This does not allow the user to define "random" external pids. (This question is directly related to my validation PR)

src/lib/components/AffiliationsField.js
ntarocco: Yes, I have auto organize import enabled. The issue is that there a few unused imports, it looks like that devs do not have ESLint well configured.
The idea is to create a ESLint module to standardize the code formatting across all repos.
Personally, I don't mind if imports are changed, but I can disabled it if you prefer. I change it only for files that I touch.

src/lib/components/Identifiers/IdentifiersField.js
ntarocco: No, it is again Prettier with ESLint with the `create-react-app` rules, which should be the most widely used in the JS community.

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+import { FieldLabel } from 'react-invenio-forms';
+import { Form, Radio } from 'semantic-ui-react';
+
+const PROVIDER_EXTERNAL = 'external';
+const UPDATE_PID_DEBOUNCE_MS = 200;
+
+/**
+ * Button component to reserve a PID.
+ */
+class BtnReservePID extends Component {
```
zzacharo, line 22: minor: Maybe we call it `ReservePIDButton` or `ReservePIDBtn`?

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+  handleClickReservePID = () => {
+    const { onPIDReserved } = this.props;
+    //PIDS-FIXME
+    console.log('TODO implement reserve PID');
+    const value = '10.1234/as7r3-234f3';
+    onPIDReserved(value);
```
zzacharo, line 28: minor: I would use in general directly the `this.props.onPIDReserved` as I think is more intuitive and says directly that the property comes from a `prop`...

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+};
+
+BtnReservePID.defaultProps = {
+  disabled: false,
+};
+
+/**
+ * Button component to unreserve a PID.
+ */
+class BtnUnreservePID extends Component {
```
zzacharo, line 58: Same for naming if you change it above :)

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+
+ManagedUnmanagedSwitch.propTypes = {
+  onManagedUnmanagedChange: PropTypes.func.isRequired,
+};
+
+/**
+ * Render identifier field and reserve/unreserve
+ * button components for managed PID.
+ */
+class ManagedIdentifierCmp extends Component {
```
zzacharo, line 139: Are we going for brief naming e.g `Component -> Cmp` or `Button -> Btn`? I am fine with it as long as we stick to something common :)

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+      <BtnUnreservePID
+        identifier={identifier}
+        onPIDUnreserved={() => onIdentifierChanged('')}
+      />
+    );
+
+    return (
+      <Form.Group>
+        <Form.Field width={8}>
+          {!hasIdentifier && (
```
zzacharo, line 168: minor: Maybe you can convert that to `hasIdentifer ? ... : ...` ?

src/lib/components/Identifiers/PIDField.js
```diff
@@ -0,0 +1,506 @@
...
+          <FieldLabel htmlFor={fieldPath} icon={pidIcon} label={pidLabel} />
+        </Form.Field>
+
+        {this.canBeManagedAndUnmanaged && (
+          <ManagedUnmanagedSwitch
+            onManagedUnmanagedChange={(isManagedSelected) => {
+              this.setState({
+                isManagedSelected: isManagedSelected,
+              });
+              const pid = this.buildPid(
```
zzacharo, line 336: Are you assuming here that the `pid` variable will be set after the state has changed above? If so, then you will need to pass a callback function to the `setState()` above :)

zzacharo (approved):

src/lib/components/Identifiers/PIDField.js
ntarocco: ESLint will complain as soon as we add it ;)

src/lib/components/Identifiers/PIDField.js
ntarocco: Went for `Cmp`, `Btn`

src/lib/components/Identifiers/PIDField.js
ntarocco: thanks for noticing, but this is on purpose, no need to have the state updated ;)

src/lib/components/Identifiers/PIDField.js
ntarocco: Fixed, ty!

src/lib/components/Identifiers/PIDField.js
ntarocco: Fixed, ty!

src/lib/components/Identifiers/PIDField.js
ntarocco: Fixed, ty!

src/lib/components/Identifiers/PIDField.js
ntarocco: Fixed, ty!

ntarocco:
If CI happy and no other comments, I will merge! Thanks for the review.

PR #264, opened 2021-04-15
files: enable/disable files via payload body

fenekku:
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/225 (in part)
- requires https://github.com/inveniosoftware/invenio-rdm-records/pull/515


src/lib/DepositController.js
```diff
@@ -200,6 +183,13 @@ export class DepositController {
       draft = response.data;
     }
 
+    // We have to save draft before we upload files, because files might have
+    // been disabled and we need to re-enable them first. We do it
+    // "stealthily" (not using saveDraft) so as to provide a nice UX.
+    const recordSerializer = store.config.recordSerializer;
+    let payload = recordSerializer.serialize(draft);
+    await this.apiClient.save(payload);
```
fenekku, line 191: yeah... we end up having to do this. We probably could be smarter about it and check if there are files already to save some calls. We might also want to take a step back when we have time to refactor some things and make it better (the checkbox could be part of formik and then we could compare against its value and the record in the redux store...)

src/lib/components/PublishButton.js
```diff
@@ -36,12 +38,9 @@ export class PublishButtonComponent extends Component {
...
     const isDisabled = (formik) => {
-      const noFilesUploaded = !numberOfFiles;
-      // Files are broken if you publish without either uploading a file
-      // or explicitely mark the draft as "Metadata only"
-      // Temporarely disable publish also in that case until the backend
-      // can validate it correctly and report back errors on partial save
-      return formik.isSubmitting || (filesEnabled && noFilesUploaded);
+      const filesMissing = filesEnabled && !numberOfFiles;
+      const hasErrors = !_isEmpty(errors);
+      return formik.isSubmitting || hasErrors || filesMissing;
```
fenekku, line 43: improving the Publish button, by disabling it if there are errors.

fenekku:
Merging this to get everything working together.

PR #266, opened 2021-04-21
files: fix file restriction display

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/776
file-protection

### DIFF
--- a/src/lib/components/AccessRightField.js
+++ b/src/lib/components/AccessRightField.js
@@ -139,7 +139,7 @@ FormikAccessRightField.defaultProps = {
 
 
 const mapStateToProps = (state) => ({
-  isMetadataOnly: !state.files.enabled,
+  isMetadataOnly: !state.deposit.record.files.enabled,
 });
 
 


src/lib/components/AccessRightField.js
```diff
@@ -139,7 +140,7 @@ FormikAccessRightField.defaultProps = {
 
 
 const mapStateToProps = (state) => ({
-  isMetadataOnly: !state.files.enabled,
+  isMetadataOnly: _isEmpty(state.files.entries),
```
fenekku, line 143: I thought so too before, but that's not the UX we want. We want that `isMetadataOnly` *mean* it is metadata-only i.e. the checkbox for metadata-only has been checked: https://github.com/inveniosoftware/invenio-app-rdm/issues/740 . Files protection should still be shown if no files but not metadata-only.

I think this should do it:

```suggestion
  isMetadataOnly: !state.deposit.record.files.enabled,
```

jrcastro2: It makes sense, your suggestion works flawless, thanks!

fenekku (approved):
Yup :smile:

PR #270, opened 2021-04-23
deposit form: Fix disabling of publish button

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/788

public-button

### DIFF
--- a/src/lib/components/PublishButton.js
+++ b/src/lib/components/PublishButton.js
@@ -15,6 +15,13 @@ import { submitAction } from '../state/actions';
 import { FORM_PUBLISHING } from '../state/types';
 import { toCapitalCase } from '../utils';
 
+const requiredFieldsForValidDraft = [
+  'title',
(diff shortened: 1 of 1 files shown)


src/lib/components/PublishButton.js
```diff
@@ -15,6 +15,13 @@ import { submitAction } from '../state/actions';
 import { FORM_PUBLISHING } from '../state/types';
 import { toCapitalCase } from '../utils';
 
+const requiredFieldsForValidDraft = [
+  'title',
+  'resource_type',
+  'publication_date',
+  'creators',
+];
+
```
jrcastro2, line 24: I don't like to have the required fields hardcoded here ... but could't find a better solution.
Is there another way of knowing if the draft is valid?

ntarocco:
Fixed by: https://github.com/inveniosoftware/react-invenio-deposit/pull/444

PR #277, opened 2021-04-26
depositForm: add display of embargo errors

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/791

image

### DIFF
--- a/src/lib/components/FormFeedback.js
+++ b/src/lib/components/FormFeedback.js
@@ -17,12 +17,11 @@ import {
 } from '../state/types';
 import { leafTraverse } from '../utils';
 
-
 const defaultLabels = {
   'files.enabled': 'Files',
   'metadata.resource_type': 'Resource type',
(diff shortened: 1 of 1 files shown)


src/lib/components/FormFeedback.js
```diff
@@ -35,14 +34,15 @@ const defaultLabels = {
   'metadata.version': 'Version',
   'metadata.publisher': 'Publisher',
   'metadata.related_identifiers': 'Related works',
-}
+  'access.embargo.until': 'Embargo until',
```
fenekku, line 37: We are slightly breaking the pattern of only showing the top level label here (other fields are not as specific)... which is fine for now (we shouldn't really care about perfect consistency). But we should keep an eye out to see if that devolved over the longer term.

fenekku (approved):
Let's go with that and keep an eye on it. (I am guessing prettier applied the various cosmetic changes...)

Issue #281, opened 2021-04-27
Clicking on "Save Draft" should jump viewport to top of the page in order to see the form feedback.


fenekku:
As a depositor, when I fill out the form, I typically scroll down (end up further down the page). Thankfully I can click on Save Draft button wherever I am because the sidebar follows along. However, the result of that action (Successful or Errors) is not immediately visible. It is only shown at the top of the page.

An easy fix is to jump the viewport to the top of the page. A more involved fix, includes delving back into the sticky feedback form.

PR #290, opened 2021-05-10
SaveButton: scroll to top after saved

fenekku:
**Edit May 12 2021**
Thanks to @zzacharo + @ntarocco : now using `scrollTo` (I was looking at `scroll` before https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll and lamenting the Safari lack of support :facepalm: ). Though it would be good for someone with Safari to doublecheck it. (below is new video on Firefox). No other dependencies.

- closes https://github.com/inveniosoftware/react-invenio-deposit/issues/281

https://user-images.githubusercontent.com/1932651/117978899-59a09480-b300-11eb-8074-2b935e35ebfb.mp4

### DIFF
--- a/src/lib/components/SaveButton.js
+++ b/src/lib/components/SaveButton.js
@@ -20,11 +20,20 @@ export class SaveButtonComponent extends Component {
   render() {
     const { formState, saveClick, ...uiProps } = this.props;
 
+    function handleClick(e, formik) {
+      saveClick(e, formik);
+      window.scrollTo({
+        top: 0,
+        left: 0,
+        behavior: 'smooth'
+      });
+    }
+
     return (
       <ActionButton
         isDisabled={this.isDisabled}
         name="save"
-        onClick={saveClick}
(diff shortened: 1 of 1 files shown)


src/lib/components/SaveButton.js
```diff
@@ -20,11 +21,16 @@ export class SaveButtonComponent extends Component {
   render() {
     const { formState, saveClick, ...uiProps } = this.props;
 
+    function handleClick(e, formik) {
+      saveClick(e, formik);
+      scroll.scrollToTop();
```
zzacharo, line 26: what about using native scrolling e.g 
```javascript
constructor(props) {
    super(props)
    this.myRef = React.createRef()   // Create a ref object 
}

componentDidMount() {
  this.myRef.current.scrollTo(0, 0);
}

render() {
    return <div ref={this.myRef}></div> 
}   // attach the ref property to a dom element
```

zzacharo (approved):

PR #297, opened 2021-06-01
deposit form: fix serialization

jrcastro2:
* Adds serialization of resource_type vocabulary
  of related_identifiers
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/864

### DIFF
--- a/src/lib/fields/ArrayFieldWithVocabulary.js
+++ b/src/lib/fields/ArrayFieldWithVocabulary.js
@@ -1,0 +1,77 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2020 CERN.
+// Copyright (C) 2020 Northwestern University.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import _isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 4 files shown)


src/lib/fields/ArrayFieldWithVocabulary.js
```diff
@@ -0,0 +1,77 @@
...
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import _isEmpty from 'lodash/isEmpty';
+import _get from 'lodash/get';
+import _set from 'lodash/set';
+import _cloneDeep from 'lodash/cloneDeep';
+
+export class ArrayFieldWithVocabulary {
```
fenekku, line 13: I don't think this is the abstraction we will keep since e.g. Additional titles has 2 vocabularies inside it (title type and language). A composable approach like marshmallow is what I have in mind for the future of the DepositRecordserializater (`field.List(field.RelatedIdentifier))`). 

For now let's keep it to clear the bugs of the release. New issue: https://github.com/inveniosoftware/react-invenio-deposit/issues/298

fenekku (approved):
We will return to the abstraction (as mentioned), but this works. Thanks!

PR #301, opened 2021-06-07
deposit: Fix lifted embargo display

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/400 

https://user-images.githubusercontent.com/15194802/121338295-a8e8dd80-c91d-11eb-8a14-9b4a85b9f942.mp4

### DIFF
--- a/src/lib/components/Access/utils.js
+++ b/src/lib/components/Access/utils.js
@@ -6,19 +6,18 @@
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 import React from 'react';
+import _isEmpty from 'lodash/isEmpty';
+
 import { DateTime } from 'luxon';
 import { Divider, Icon, List, Message } from 'semantic-ui-react';
-import {
-  TextAreaField,
-} from 'react-invenio-forms';
+import { TextAreaField } from 'react-invenio-forms';
 
 import { EmbargoState } from './Embargo';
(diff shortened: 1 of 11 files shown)


src/lib/components/Access/utils.js
```diff
@@ -44,27 +39,27 @@ export function filesSection(filesStyle, filesContent) {
...
-
-export function embargoSection(embargo) {
-  const fmtDate = (
-    embargo.date
-    ? DateTime.fromISO(embargo.date)
-      .toLocaleString(DateTime.DATE_FULL) // e.g. June 21, 2021
-    : "???"
-  );
+export function embargoSection(initialAccessValues, embargo) {
+  const fmtDate = initialAccessValues?.embargo?.until
```
ntarocco, line 55: minor: it would be nice to check how babel will convert the `?` operator: if it is adding `if`s around, I would do it once at the beginning and then avoid it later on...

ntarocco (approved):
If it happens that you re-test it, please add a screenshot! :)

src/lib/components/Access/utils.js
jrcastro2: Hmm, we trully only need to check for `initialAccessValues.embargo` since `initialAccessValues` should not be null, added it just in case, but will remove it since we only need to check for the `embargo`

jrcastro2:
> If it happens that you re-test it, please add a screenshot! :)

Yes, added, thanks!

ntarocco (approved):

PR #307, opened 2021-06-22
i18n: adds translation support

mb-wali:
related issue inveniosoftware/react-invenio-forms/issues/6

https://user-images.githubusercontent.com/44528277/122896820-bf455f00-d349-11eb-9a72-1fb4d4d12ba9.mp4

### Dependencies
*  `i18next` 
* `i18next-browser-languagedetector` (detect user language)
* `i18next-scanner` (extract translation keys/values)
* `i18next-conv` (converts files from gettext (.mo/.po) to i18next's json format and vice versa)

### Mark for translation
import created instance in your package.
`import i18next from '../i18next';`

Mark for translation
``` diff
--- <List.Item>Storage available</List.Item>
+++ <List.Item>{i18next.t('Storage available')}</List.Item>
```

### Extract translations
To scan your code, extract translation keys/values, and merge them into i18n resource files.
```npm
npm run i18n-scan
```
### Create/update **.pot** file, this file is pushed to transifex 
converts source language file from `JSON` to `.pot`
```npm
npm run i18n-conv-json
```

### Convert/update .po files to json.
once translation is pulled from transifex, the format is `.po`, in order to convert these files to JSON format run following command.
```npm
npm run i18n-conv-po --lang="<lang>"
```
### Convert/update all the .po files for all available languages
```npm
npm run i18n-conv-po-all
```

### DIFF
--- a/i18next-scanner.config.js
+++ b/i18next-scanner.config.js
@@ -1,0 +1,69 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2021 Graz University of Technology.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+const fs = require('fs');
+const chalk = require('chalk');
+const { languages } = require('./package').config;
+
+// list of func used to 
+// mark the strings for translation
+const funcList = ['i18next.t', 'i18n.t', 't'];
+
+module.exports = {
+    options: {
+        debug: true,
+        browserLanguageDetection: true,
+        func: {
+            list: funcList,
+            extensions: ['.js', '.jsx']
+        },
+        trans: false, // Enable for using Trans component
+        lngs: languages,
+        ns: [
+            // file name (.json)
+            'translations',
+        ],
+        defaultLng: 'en',
+        defaultNs: 'translations',
+        resource: {
+            // The path where resources get loaded from. Relative to current working directory. 
+            loadPath: 'src/lib/translations/{{lng}}/{{ns}}.json',
+
+            // The path to store resources.
+            savePath: 'src/lib/translations/{{lng}}/{{ns}}.json',
+
+            jsonIndent: 2,
+            lineEnding: '\n'
+        },
+        nsSeparator: false, // namespace separator
+
+        //Set to false to disable key separator
+        // if you prefer having keys as the fallback for translation (e.g. gettext). 
+        keySeparator: false,
+    },
+
+    // func is provided
+    // (if needed)for more custom extractions.
+    transform: function customTransform(file, enc, done) {
+        "use strict";
+        const parser = this.parser;
+        const content = fs.readFileSync(file.path, enc);
+        let count = 0;
+
+        parser.parseFuncFromString(content, { list: funcList }, (key, options) => {
+            parser.set(key, Object.assign({}, options, {
+                nsSeparator: false,
+                keySeparator: false
+            }));
+            ++count;
+        });
+
+        if (count > 0) {
+            console.log(`i18next-scanner: count=${chalk.cyan(count)}, file=${chalk.yellow(JSON.stringify(file.relative))}`);
+        }
+        done();
+    }
+};
--- a/package.json
+++ b/package.json
@@ -8,6 +8,13 @@
   "files": [
     "dist"
   ],
+  "config": {
+    "languages": [
+      "en",
+      "de",
+      "tr"
+    ]
+  },
   "scripts": {
     "start": "react-scripts start",
     "build": "rimraf dist && NODE_ENV=production rollup -c",
@@ -17,6 +24,10 @@
     "postlink-dist": "cd dist && rm -rf node_modules",
     "unlink-dist": "cd dist && npm unlink && rm package*",
     "watch": "NODE_ENV=development rollup --watch -c",
+    "i18n-scan": "i18next-scanner --config i18next-scanner.config.js src/**/*.{js,jsx}",
+    "i18n-conv-json": "i18next-conv -l en -s ./src/lib/translations/en/translations.json -t ./src/lib/translations/translations.pot",
+    "i18n-conv-po": "i18next-conv -l $npm_config_lang -s ./src/lib/translations/$npm_config_lang/messages.po -t ./src/lib/translations/$npm_config_lang/translations.json",
+    "i18n-conv-po-all": "node ./getTextConverter.js",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint src/**/*.js",
@@ -36,7 +47,9 @@
     "lodash": "^4.17.15",
     "semantic-ui-react": "^0.88.0",
     "semantic-ui-css": "^2.4.1",
-    "@babel/runtime": "^7.9.0"
+    "@babel/runtime": "^7.9.0",
+    "i18next": "^20.3.1",
+    "i18next-browser-languagedetector": "^6.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.5.5",
@@ -44,6 +57,7 @@
     "@ckeditor/ckeditor5-react": "^2.1.0",
     "@rollup/plugin-babel": "^5.0.0",
     "@rollup/plugin-commonjs": "^11.1.0",
+    "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^7.1.3",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.5.0",
@@ -55,6 +69,8 @@
     "enzyme-to-json": "^3.4.3",
     "expect": "^26.0.1",
     "formik": "^2.1.4",
+    "i18next-conv": "^10.2.0",
+    "i18next-scanner": "^3.0.0",
     "json": "^9.0.6",
     "lodash": "^4.17.15",
     "prettier": "^2.0.5",
--- a/src/lib/i18next.js
+++ b/src/lib/i18next.js
@@ -1,0 +1,49 @@
+// This file is part of React-Invenio-Deposit
+// Copyright (C) 2021 Graz University of Technology.
+//
+// React-Invenio-Deposit is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import i18n from 'i18next';
+
+import TRANSLATIONS_DE from './translations/de/translations.json';
+import TRANSLATIONS_EN from './translations/en/translations.json';
+import TRANSLATIONS_TR from './translations/tr/translations.json';
+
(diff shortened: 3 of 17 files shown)


src/lib/i18next.js
```diff
@@ -0,0 +1,60 @@
...
+// import { TRANSLATIONS_DE } from './translations/de/translations';
+// import { TRANSLATIONS_EN } from './translations/en/translations';
+
+// import translations json
+// (Note that you need @rollup/plugin-json to import JSON files)
+// rollup.config.js
+//  - import json from "@rollup/plugin-json";
+//  - json(),
+import TRANSLATIONS_DE from './translations/de/translations.json';
+import TRANSLATIONS_EN from './translations/en/translations.json';
```
mb-wali, line 22: For now, it supports both - `.js` & `.json` format of translation files - up for discussion which one could we use in order to push and pull to transifex. also there are plugins to convert these formats.

src/lib/i18next.js
```diff
@@ -0,0 +1,60 @@
...
+      en: {
+        translation: TRANSLATIONS_EN,
+      },
+      de: {
+        translation: TRANSLATIONS_DE,
+      },
+    },
+    // specify language detection order
+    detection: {
+      order: ['htmlTag'],
```
lnielsen, line 50: How does this detection work?  From Jinja we embed the selected user language? What happens if you change the language on the deposit form itself? We reload the page?

src/lib/i18next.js
```diff
@@ -0,0 +1,60 @@
...
+  .use(LanguageDetector)
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    fallbackLng: 'en',
+    debug: true,
+    resources: {
+      en: {
```
lnielsen, line 41: I assume we would have to embed the translations of all languages right? I guess it's not too big an overhead, compared to having to find a way to only include the once an instance are using.

src/lib/i18next.js
lnielsen: You mean transfix can deal with both formats?

src/lib/translations/en/translations.js
```diff
@@ -0,0 +1,6 @@
+export const TRANSLATIONS_EN = {
+    "metadata.title": "Title",
+    "form.title.new.deposit": "New Upload",
```
lnielsen, line 3:
```suggestion
    "form.title.new.deposit": "New upload",
```

lnielsen: ditto in the other places.

src/lib/i18next.js
mb-wali: not exactly these two formats but for JSON it should be `ICU` and for javascript it should be `RequireJS`, and i18next does support most of them, and for the one, it does not, it has plugins to do the conversion. it depends on the need of us which format do we want to use - and it would be nice to create an account for this module and test the actual pull and push.

**related links**: [transifex formats](https://docs.transifex.com/formats/introduction) &  [i18next formats](https://www.i18next.com/overview/plugins-and-utils#i-18-n-formats)

src/lib/i18next.js
mb-wali: This order detects the change in the `htmlTag (add html language tag <html lang="LANGUAGE" ...)`, there are several orders available I can investigate for our need. https://github.com/i18next/i18next-browser-languageDetector#introduction

yes in this detection order - if you change the language on the deposit form it will reload the page.

src/lib/i18next.js
mb-wali: i18next - supports different approaches to add the resources(translations), see also [load-translations](https://www.i18next.com/how-to/add-or-load-translations), in this example I used something similar to python modules.

package.json
```diff
@@ -22,7 +22,12 @@
     "lint": "eslint src/**/*.js",
     "format": "prettier --config ./.prettierrc --ignore-path ./.prettierignore --write \"src/**/*.js\""
   },
-  "dependencies": {},
+  "dependencies": {
+    "@rollup/plugin-json": "^4.1.0",
```
ntarocco, line 26: This should go to `devDeps`

package.json
```diff
@@ -22,7 +22,12 @@
...
+    "i18next": "^20.3.1",
```
ntarocco, line 27: These 3 should go to `peerDeps`

src/lib/index.js
```diff
@@ -13,3 +14,5 @@ export { DepositRecordSerializer } from './DepositRecordSerializer';
 export { connect } from './connect';
 export * from './dom';
 export * from './components';
+// import i18n (needs to be bundled ;))
```
ntarocco, line 17:
```suggestion

```

ntarocco (commented):
We have discussed with @zzacharo and it looks a nice implementation, it is maybe missing a little bit more "plumbing".

To extract all the translations, it looks like that we can use [18next-scanner](https://github.com/i18next/i18next-scanner) (to be tested!!!). It could be added as a command in the `package.json`.
Then we will have to document how to upload the extracted keys to Transifex, and we might need some config file here for the Transifex integration.

At the end, when we download the `.po` catalog from Transifex with the translations, we found that we could use [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter), again to be tested and integrated in the `package.json`.

What do you think?

mb-wali:
> We have discussed with @zzacharo and it looks a nice implementation, it is maybe missing a little bit more "plumbing".
> 
> To extract all the translations, it looks like that we can use [18next-scanner](https://github.com/i18next/i18next-scanner) (to be tested!!!). It could be added as a command in the `package.json`.
> Then we will have to document how to upload the extracted keys to Transifex, and we might need some config file here for the Transifex integration.
> 
> At the end, when we download the `.po` catalog from Transifex with the translations, we found that we could use [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter), again to be tested and integrated in the `package.json`.
> 
> What do you think?

I like it, sounds like a plan.   I will start digging into this. Thanks

i18next-scanner.config.js
```diff
@@ -0,0 +1,61 @@
...
+const fs = require('fs');
+const chalk = require('chalk');
+
+module.exports = {
+    options: {
+        debug: true,
+        // removeUnusedKeys: true, // removing unused keys
+        browserLanguageDetection: true,
+        func: {
+            list: ['i18next.t', 'i18n.t', 't'],
```
zzacharo, line 16: What is this list defining? The way you import and mark the strings for translation?

i18next-scanner.config.js
```diff
@@ -0,0 +1,61 @@
...
+        // if you prefer having keys as the fallback for translation (e.g. gettext). 
+        keySeparator: false,
+    },
+    transform: function customTransform(file, enc, done) {
+        "use strict";
+        const parser = this.parser;
+        const content = fs.readFileSync(file.path, enc);
+        let count = 0;
+
+        parser.parseFuncFromString(content, { list: ['i18next.t', 'i18n.t', 't'] }, (key, options) => {
```
zzacharo, line 49: Is that list the same as defined above? If so, maybe we should define it as a const and reuse it here?

package.json
```diff
@@ -17,6 +17,8 @@
     "postlink-dist": "cd dist && rm -rf node_modules",
     "unlink-dist": "cd dist && npm unlink && rm package*",
     "watch": "NODE_ENV=development rollup --watch -c",
+    "i18n-scan": "i18next-scanner --config i18next-scanner.config.js src/**/*.{js,jsx}",
+    "i18n-conv-json": "i18next-conv -l en -s ./src/lib/translations/en/translations.json -t ./src/lib/translations/translations.pot",
```
zzacharo, line 21: Is this the conversion from the `.pot` file to json? If so, the `-l` parameter should come as an argument for each language no?

src/lib/i18next.js
```diff
@@ -0,0 +1,44 @@
...
+import TRANSLATIONS_DE from './translations/de/translations.json';
+import TRANSLATIONS_EN from './translations/en/translations.json';
+import TRANSLATIONS_TR from './translations/tr/translations.json';
+
+import LanguageDetector from 'i18next-browser-languagedetector';
+
+i18next
+  // detect user language
+  .use(LanguageDetector)
+  .init({
```
zzacharo, line 18: Are the translations here part of the production bundle? There is no request happening once the js is loaded right?

mb-wali: right - they are bundled with the help of rollup, after production bundle they are found in the `/dist/cjs/index.js` & `/dist/esm/index.js`.

i18next-scanner.config.js
mb-wali: yes, list of func name you import and mark the strings for translation?

i18next-scanner.config.js
```diff
@@ -0,0 +1,71 @@
...
+
+const fs = require('fs');
+const chalk = require('chalk');
+
+// list of func used to 
+// mark the strings for translation
+const funcList = ['i18next.t', 'i18n.t', 't'];
+
+// available languages
+const languages = ['en','de', 'tr'];
```
zzacharo, line 15: That can work for now but we should think what are the needed steps to add a new language to this repository. Maybe we could create an issue for now explaining the process...

i18next-scanner.config.js
```diff
@@ -0,0 +1,71 @@
...
+            // file name (.json)
+            'translations',
+        ],
+        defaultLng: 'en',
+        defaultNs: 'translations',
+        resource: {
+            // The path where resources get loaded from. Relative to current working directory. 
+            loadPath: 'src/lib/translations/{{lng}}/{{ns}}.json',
+
+            // The path to store resources. Relative to the path specified by `gulp.dest(path)`.
```
zzacharo, line 37: The `gulp.dest` comment is not relevant here...right?

package.json
```diff
@@ -17,6 +17,9 @@
...
+    "i18n-conv-po": "i18next-conv -l $npm_config_lang -s ./src/lib/translations/$npm_config_lang/messages.po -t ./src/lib/translations/$npm_config_lang/translations.json",
```
zzacharo, line 22: Maybe we will need to have `i18n-conv-po-all` function to compile all the `.po` files for all available languages...Maybe the configured languages could be set here as a config variable https://docs.npmjs.com/cli/v7/configuring-npm/package-json#config

zzacharo (commented):
It looks very good @mb-wali !!! Just some small comments but the flow seem neat!

package.json
```diff
@@ -17,6 +24,10 @@
...
+    "i18n-conv-po-all": "node ./getTextConverter.js",
```
zzacharo, line 30: @mb-wali what about merging the `i18n-conv-po` and this command by simply adding an extra argument e.g `--all` to comply also with what `tx` command is using for pulling all languages? I can approve this PR if you think this is a lot of work to be done now and we can open an issue for that :)

PR #309, opened 2021-06-24
creatibutors: fix modal bug

jrcastro2:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/960
* reopens modal when saving and adding
  another one to unmount and mount again
  the components and reset their state

### Before
https://user-images.githubusercontent.com/15194802/123294630-514d9300-d515-11eb-889e-b479088ccb50.mp4

### After
https://user-images.githubusercontent.com/15194802/123294634-527ec000-d515-11eb-9500-b680db6659a2.mp4

### DIFF
--- a/src/lib/components/Creatibutors/CreatibutorsModal.js
+++ b/src/lib/components/Creatibutors/CreatibutorsModal.js
@@ -171,6 +171,10 @@ export class CreatibutorsModal extends Component {
     formikBag.resetForm();
     switch (this.state.action) {
       case 'saveAndContinue':
+        // Needed to close and open the modal to reset the internal
+        // state of the cmp inside the modal
+        this.closeModal();
+        this.openModal();
         this.changeContent();
         break;
       case 'saveAndClose':


src/lib/components/Creatibutors/CreatibutorsModal.js
```diff
@@ -171,6 +171,8 @@ export class CreatibutorsModal extends Component {
     formikBag.resetForm();
     switch (this.state.action) {
       case 'saveAndContinue':
+        this.closeModal();
```
ntarocco, line 174: Can we do instead:
```
formikBag.resetForm();
this.focusInput();
```
?

ntarocco (commented):
Nothing do with your PR, but there are multiple issues with this modal:
1. it is added in a for-loop, and it should be not. The modal should be one only.
2. it would be nice to improve slightly the function to have the `Save and add another` btn disabled while it has the label `Added...` (with the 3 dots?):
```
changeContent = () => {
    this.setState({ saveAndContinueLabel: 'Added...', saveAndContinueEnabled: false });
    // change in 2 sec
    setTimeout(() => {
      this.setState({ saveAndContinueLabel: 'Save and add another', saveAndContinueEnabled: true });
    }, 2000);
  };
```
On top of that, to be even more correct, it should be changed to:
```
this.setState({ ... }, () => setTimeout(...));
```
It works without because of the setTimeout, but ideally we should trigger a sequential state change *after* the first is really mutated.
3. The `setTimeout` might trigger memory leaks. If you close the modal in between those 2 seconds, and for some reasons the cmp is unmounted, React will try to change the state for a cmp that does not exist anymore.

src/lib/components/Creatibutors/CreatibutorsModal.js
jrcastro2: This isn't working unfortunately mainly because the problem is not in the `formik` values, the issue is with the [TextFields](https://github.com/inveniosoftware/react-invenio-forms/blob/master/src/lib/TextField.js) because it manages the state internally and only updates the formik values [onBlur](https://github.com/inveniosoftware/react-invenio-forms/blob/master/src/lib/TextField.js#L41-L47). If we keep the modal open and only reset the formik values the internal state of `TextField` won't reset and keep the previous values. Maybe another way to do it is to check if formik values were submitted evey time we update the [TextField](https://github.com/inveniosoftware/react-invenio-forms/blob/master/src/lib/TextField.js) and if so, reset the state, WDYT?

ntarocco: Let's keep it like this for the time being. Modals should be refactored to avoid this too and handle the state correctly.

I would simply add a comment to explain why this is needed because otherwise it looks weird, push and merge :)

fenekku (approved):
I agree with what was said. This one can be merged.

PR #310, opened 2021-06-24
pids: remove hardcoded unmanaged provider name

### LINKED ISSUE #285: pids: deposit form error validation
**Identifier value**
When sending a wrong value for an identifier the API will answer with an error for the field `pids`. This error is:

- Does not appear in the list of fields shown at the top of the deposit form.
- Does not highlight in red the `identifiers` field.

Note that due to Marshmallow validation flows, the API reports errors for the gloabl `pids` field. There is only one use case when the full `pids.{scheme}.value.identifier` is given back; which still contains an extra `value` in the middle. More info in this PR https://github.com/inveniosoftware/invenio-rdm-records/pull/578 (See also comments in the code).

**Alternate identifier**
Adding an invalid identifier in alternate identifiers reports “Unknown field”

ntarocco:
* refers to inveniosoftware/invenio-rdm-records#580
 * closes #285
 * closes https://github.com/inveniosoftware/invenio-rdm-records/issues/763

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13270,9 +13270,9 @@
       "dev": true
     },
     "react-invenio-forms": {
-      "version": "0.8.0",
-      "resolved": "https://registry.npmjs.org/react-invenio-forms/-/react-invenio-forms-0.8.0.tgz",
-      "integrity": "sha512-Js8/HeJfDWhj1exW6weQe254P67cUs4OJgxZS2A1HYNOFFUftXbZCvS7z5Se04qm2PojNniZr6ULF7Sa/pextA==",
+      "version": "0.8.1",
+      "resolved": "https://registry.npmjs.org/react-invenio-forms/-/react-invenio-forms-0.8.1.tgz",
+      "integrity": "sha512-zKBZHa4eB0aGYr08HSpCB7l6oQEYbP5Zx16DiNQkH0C1YfL26a/cF8fdzGdCsSKy7WuV+CJkPAQB5SzbdPwIHA==",
       "dev": true
     },
     "react-is": {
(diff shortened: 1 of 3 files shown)


ntarocco:
Screenshot 2021-06-24 at 17 41 15

ntarocco:
@jrcastro2 can you please review this PR?

jrcastro2 (commented):
Overall LGTM.
There is just a small bug that needs to be fixed, when saving a wrong DOI and then clicking the "No" radio button for DOI I would expect to save it without a DOI, instead it still saves the previous DOI and the error message is still visible.

https://user-images.githubusercontent.com/15194802/123663108-5e7cc180-d836-11eb-84c1-6dbb1d20fb43.mp4

ntarocco:
> Overall LGTM.
> There is just a small bug that needs to be fixed, when saving a wrong DOI and then clicking the "No" radio button for DOI I would expect to save it without a DOI, instead it still saves the previous DOI and the error message is still visible.
> pid-bug.mp4

Thanks a lot for taking the time of checking it out and testing it. I have simply removed the error from the `Managed` part because it does not make sense in this case. We will see if we need errors there, but it should not be the case.

jrcastro2 (approved):
Now the error message doesn't appear but it seems that there are still some bugs concerning the DOI, I presume this can be fixed in a separated issue:
- If you paste a DOI and click NO it still saves the previous one (I would expect nothing to be saved)
- If you paste a DOI, then delete it and try to save it, the UI breaks.

https://user-images.githubusercontent.com/15194802/124292825-466ab200-db56-11eb-9f06-996c2b6106bc.mp4

PR #315, opened 2021-06-29
fields: refactor array vocabulary field into schema field

ppanero:


src/lib/DepositRecordSerializer.test.js
```diff
@@ -324,6 +319,7 @@ describe('DepositRecordSerializer', () => {
         metadata: {
           contributors: [
             {
+              affiliations: [],
```
ppanero, line 322: It is a bit unclear what's expected with the "default" (de)serilization. In a test above, all fields are given the default value if not present, e.g. for languages. Therefore, I am assuming the same happens here for the inner fields of a creatibutor. We should remove these when sending to the API (I think we are doing it already tho).

src/lib/fields/Field.js
```diff
@@ -28,6 +28,27 @@ export class Field {
...
+    //                 (re-using deserialize has some caveats)
+    //                 Form/Error UX is tackled in next sprint and this is good
+    //                 enough for now.
+    for (let e of errors) {
+      _set(deserializedErrors, e.field, e.messages.join(' '));
+    }
+
+    return deserializedErrors;
+  }
+
```
jrcastro2, line 51: I don't think we need this, or do we?

src/lib/fields/VocabularyField.js
```diff
@@ -12,40 +12,44 @@ import _cloneDeep from 'lodash/cloneDeep';
...
+    serializedDefault = null,
+    labelField = 'title',
+  }) {
+    super({fieldpath, deserializedDefault, serializedDefault})
+    this.labelField = labelField
+  }
+
   deserialize(record) {
-    let fieldValue = _get(record, this.fieldpath, this.deserializedDefault);
+    console.error("before de", record)
```
jrcastro2, line 26:
```suggestion
```

jrcastro2 (approved):
LGTM!
Great job, the new `SchemaField` it's a major improvement in terms of readability and consistency, thanks a lot! :rocket: 
(just a couple of minor comments)

src/lib/fields/Field.js
ppanero: I think "maybe" since we are deserializing schemas, so errors have correct field paths... but let's remove and see in the future. We need to put some more work on error de/serialization

src/lib/fields/VocabularyField.js
ppanero: oops... *hides in shame*

PR #333, opened 2021-07-14
i18n: extarct strings for translation

mb-wali:


src/lib/components/PublicationDateField.js
```diff
@@ -18,10 +19,10 @@ export class PublicationDateField extends Component {
     return (
       <TextField
         fieldPath={fieldPath}
-        helpText={
-          'In case your upload was already published elsewhere, \
-          please use the date of the first publication. Format: YYYY-MM-DD, YYYY-MM, or YYYY. For intervals use DATE/DATE, e.g. 1939/1945.'
-        }
+        helpText={i18next.t(
+          'In case your upload was already published elsewhere, please use the date of the first publication. Format: YYYY-MM-DD, YYYY-MM, or YYYY. For intervals use DATE/DATE, e.g. 1939/1945.',
+          { keySeparator: '>', nsSeparator: '|' }
```
mb-wali, line 24: this was required to escape the colon: see also https://github.com/i18next/i18next/issues/361

zzacharo: The keyseparator option here https://www.i18next.com/overview/configuration-options#misc says that we can set that globally to false if we use a flat json which is our case I think :)

zzacharo (approved):

PR #335, opened 2021-07-15
licenses: Adapt to vocab structure

jrcastro2:
* New (de)serializer for this field

* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/718
* Depends https://github.com/inveniosoftware/invenio-rdm-records/pull/736

### DIFF
--- a/src/lib/components/License/LicenseField.js
+++ b/src/lib/components/License/LicenseField.js
@@ -24,6 +24,7 @@ class LicenseFieldForm extends Component {
       label,
       labelIcon,
       fieldPath,
+      uiFieldPath,
       form: { values, errors },
       move: formikArrayMove,
       push: formikArrayPush,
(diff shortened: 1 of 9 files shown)


src/lib/components/License/LicenseField.js
```diff
@@ -44,10 +71,29 @@ class LicenseFieldForm extends Component {
               const arrayPath = fieldPath;
               const indexPath = index;
               const key = `${arrayPath}.${indexPath}`;
+              const uiKey = `${uiFieldPath}.${indexPath}`;
               const licenseType = value.id ? 'standard' : 'custom';
-              const description = getIn(values, `${key}.description`, '');
-              const link = getIn(values, `${key}.link`, '');
-              const title = getIn(values, `${key}.title`, '');
+              const description = getIn(
```
jrcastro2, line 76: Here now we need to check for the UIobject for any already loaded license but we need to rely in the value inside "metadata" for the added values from the interface, I cannot find a solution to only use the UIObject.

src/lib/fields/VocabularyField.js
```diff
@@ -87,3 +87,106 @@ export class AllowAdditionsVocabularyField extends VocabularyField {
...
+   * @param {String} defaultLocale - The default locale
+   * @returns
+   */
+  deserialize(record, defaultLocale) {
+
+    const fieldValue = _get(record, this.fieldpath, this.deserializedDefault);
+    const _deserialize = (value) => {
+      if ('id' in value) {
+        if (typeof value.title === 'string') {
+          // Needed in case we pass a default value
```
jrcastro2, line 122: Needed for [this](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L495-L509). Ideally we would just add the id as default and retrieve the value in the UIObject but it's not really working because we don't have a record created and we only serialize it when we edit (check [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/deposits.py#L224-L225)).

PR #336, opened 2021-07-15
i18n: mark strings for translation

mb-wali:


src/lib/components/Access/Embargoed.js
```diff
@@ -22,33 +29,45 @@ export class Embargoed {
...
-      : "???"
-    );
+    const fmtDate = this.embargo.date
+      ? DateTime.fromISO(this.embargo.date).toLocaleString(DateTime.DATE_FULL) // e.g. June 21, 2021
+      : '???';
 
-    const text = <>On <b>{fmtDate}</b> the record and the files will automatically be made publicly accessible. Until then, the record and the files can <b>only</b> be accessed by <b>users specified</b> in the permissions.</>;
+    const text = (
+      <>
+        {/* TODO: use of Trans Component for jsx translations */}
```
mb-wali, line 50: i have added a `TODO` just in case if we are not happy in the future translating `jsx` format like this.

zzacharo: Can't you use the string interpolation here? How does it look like the extracted string if you include the html elements?

zzacharo (approved):

PR #338, opened 2021-07-15
global: fix list options sorting

ppanero:
- closes https://github.com/inveniosoftware/react-invenio-deposit/issues/337

### DIFF
--- a/src/lib/components/AdditionalDescriptionsField.js
+++ b/src/lib/components/AdditionalDescriptionsField.js
@@ -14,6 +14,7 @@ import { ArrayField, SelectField, RichInputField } from 'react-invenio-forms';
 import { emptyAdditionalDescription } from '../record';
 import { LanguagesField } from './LanguagesField';
 import { i18next } from '@translations/i18next';
+import { sortOptions } from '../utils';
 
 export class AdditionalDescriptionsField extends Component {
   render() {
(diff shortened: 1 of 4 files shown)


src/lib/components/AdditionalDescriptionsField.js
```diff
@@ -16,6 +16,10 @@ import { LanguagesField } from './LanguagesField';
 import { i18next } from '../i18next';
 
 export class AdditionalDescriptionsField extends Component {
+  sortOptions = (options) => {
```
ppanero, line 19: This function is now repeated over 3 different components... can we make it generic somehow? It wouldn't make sense in `Field`... Ideas?

fenekku: Until there is a better place, we can place it here: https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/utils.js 

```js
function sortOptions(options) {
    return options.sort((o1, o2) => o1.text.localeCompare(o2.text));
}
```

fenekku (approved):

src/lib/components/AdditionalDescriptionsField.js
ppanero: Thanks!

PR #345, opened 2021-07-19
i18n: removed strings generated with scanner

mb-wali:


mb-wali:
i18next scanner - has some issue while extracting messages only if the strings are marked with `Trans` component.
the issue is that it passes the key to values for other languages rather than only to eng language.

zzacharo (commented):
@mb-wali what about adding here the https://github.com/i18next/react-i18next-gitbook/blob/master/latest/i18nextprovider.md#when-to-use as it this is the component library and leave the global context i.e the default one, in the https://github.com/inveniosoftware/invenio-app-rdm/pull/978 ?

PR #347, opened 2021-07-19
validation: enable error display on create

jrcastro2:
* Remove "enableReinitialize" formik prop

* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/965

https://user-images.githubusercontent.com/15194802/126297752-9267f694-e9a7-4eb3-a10e-e3b9f33d2d0c.mp4

### DIFF
--- a/src/lib/DepositBootstrap.js
+++ b/src/lib/DepositBootstrap.js
@@ -30,8 +30,9 @@ class DepositBootstrapComponent extends Component {
       <BaseForm
         onSubmit={this.props.submitFormData}
         formik={{
-          enableReinitialize: true,
+          enableReinitialize: true, // Needed for files
           initialValues: this.props.record,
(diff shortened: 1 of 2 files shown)


src/lib/DepositBootstrap.js
```diff
@@ -30,7 +30,6 @@ class DepositBootstrapComponent extends Component {
       <BaseForm
         onSubmit={this.props.submitFormData}
         formik={{
-          enableReinitialize: true,
```
jrcastro2, line 33: Is there a reason why we reset the form when `initialValues` changes?

(I tried to create/edit/publish records and works OK without it)

zzacharo (approved):
@jrcastro2 it looks fine if you test the workflow and it's working!

src/lib/DepositBootstrap.js
jrcastro2: This is needed, otherwise files are not working properly

PR #350, opened 2021-07-20
FormFeedback: remove duplicate error messages

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/964

screencapture-127-0-0-1-5000-uploads-s02s0-xc746-2021-07-20-11_40_43

### DIFF
--- a/src/lib/components/FormFeedback.js
+++ b/src/lib/components/FormFeedback.js
@@ -24,13 +24,12 @@ const defaultLabels = {
   'files.enabled': i18next.t('Files'),
   'metadata.resource_type': i18next.t('Resource type'),
   'metadata.title': i18next.t('Title'),
-  'metadata.additional_titles': i18next.t('Title'), // to display under Title label
+  'metadata.additional_titles': i18next.t('Additional titles'),
   'metadata.publication_date': i18next.t('Publication date'),
   'metadata.creators': i18next.t('Creators'),
   'metadata.contributors': i18next.t('Contributors'),
   'metadata.description': i18next.t('Description'),
-  // to display under Description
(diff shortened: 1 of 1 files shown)


src/lib/components/FormFeedback.js
```diff
@@ -57,12 +57,13 @@ class DisconnectedFormFeedback extends Component {
    * @returns String or React node
    */
   renderErrorMessages(messages) {
-    if (messages.length === 1) {
+    let uniqueMessages = messages.filter((v, i, a) => a.indexOf(v) === i);
```
fenekku, line 60: Cool, I was thinking [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) but that works :+1: .

fenekku (approved):
I think I am fine with the non-repeating error messages (although it could be argued both ways: clearer because showed 1 or less clear because not reflecting multiple errors).

**Action**: We will have the same issue with `Additional Descriptions`, so let's deal with it now.

Overall, I am happy the changes were pretty localized.

src/lib/components/FormFeedback.js
zzacharo: @jrcastro2 I would go for either Set as @fenekku suggested or at least https://lodash.com/docs/4.17.15#uniqBy . It's much clearer I think the intention of what we try to achieve here.....

jrcastro2: Yes, changed to Set, thanks!

P.S: `Additional Descriptions` added.

fenekku:
That still works for me :+1:

PR #357, opened 2021-07-20
PIDField: capitalize acronym

fenekku:


src/lib/components/Identifiers/PIDField.js
```diff
@@ -56,7 +56,7 @@ class UnreservePIDBtn extends Component {
     const { disabled, handleDiscardPID, loading, pidType } = this.props;
     return (
       <Popup
-        content={'Discard the reserved ' + pidType}
+        content={'Discard the reserved ' + pidType.toUpperCase()}
```
lnielsen, line 59: @ppanero have had similar cases on the backend to this one. We cannot assume that all PID types can be capitalized since there are some cases where it doesn't work - e.g. ``Bibcode``, ``arXiv``, and custom ones like ``CERN Report Number``.

ppanero: The labels are getting injected to the form <https://github.com/inveniosoftware/invenio-app-rdm/pull/971/files>. Maybe u can use something from there?

zzacharo:
After IRL discussions this is superseded by https://github.com/inveniosoftware/invenio-app-rdm/pull/1001

PR #392, opened 2021-09-27
Display subject w scheme

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/799

### DIFF
--- a/src/lib/fields/VocabularyField.js
+++ b/src/lib/fields/VocabularyField.js
@@ -65,26 +65,49 @@ export class VocabularyField extends Field {
 }
 
 export class AllowAdditionsVocabularyField extends VocabularyField {
-  deserialize(record) {
-    const fieldValue = _get(record, this.fieldpath, this.deserializedDefault);
-    // We deserialize the values in the format
-    // {id: 'vocab_id', <labelField>: 'vacab_name'} for controlled values
-    // and {<labelField>: 'vocab_name'} for user added entries
-    const _deserialize = (value) => ({
+  // We deserialize the values in the format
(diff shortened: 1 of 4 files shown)


src/lib/fields/VocabularyField.js
```diff
@@ -65,20 +65,23 @@ export class VocabularyField extends Field {
...
+
+  deserialize(record) {
+    const fieldValue = _get(record, this.fieldpath, this.deserializedDefault);
     let deserializedValue = null;
     if (fieldValue !== null) {
       deserializedValue = Array.isArray(fieldValue)
-        ? fieldValue.map(_deserialize)
-        : _deserialize(fieldValue);
+        ? fieldValue.map((value) => this._deserialize(value, this))
+        : this._deserialize(fieldValue, this);
```
fenekku, line 84: We don't need to pass `this` as an argument, right?

fenekku (approved):
Let's remove the `this` and get this merged :+1:

src/lib/fields/VocabularyField.js
jrcastro2: Hmm, I passed it to have access to `this.labelField` (line 74), otherwise we do not have access to it, or do we?

fenekku: We do now that `_deserialize` is a method :) . No need to specify it in the method (more like Java as opposed to Python) For more: https://javascript.info/object-methods#this-in-methods

jrcastro2: I see, thanks!

PR #393, opened 2021-10-01
pids: adapt to new backend

zzacharo:
depends on https://github.com/inveniosoftware/invenio-app-rdm/pull/1103

# Current possibilities

## Edit unpublished draft

- You can add/discard and change between external and managed doi freely
- If you have saved and external DOI then you will need to "discard" it before adding a new one. The discard action is just a save of the draft with `{pids: {doi: {}}}`

## Edit published record

- If the published record has an external DOI then you can discard it and either add a new one or reserve a managed one. **Note**: If you reserve a managed DOI in this case you won't be able to discard it even before publishing!
- If the published record has a managed DOI then you cannot change it!

### DIFF
--- a/src/lib/DepositController.js
+++ b/src/lib/DepositController.js
@@ -258,10 +258,10 @@ export class DepositController {
   /**
    * Discard a previously reserved PID
    */
-  discardPID = async (draft, pidType, { formik, store }) => {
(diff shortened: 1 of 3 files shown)


zzacharo:
Closing as it quite outdated

PR #404, opened 2021-12-18
ORCID icon: size fixed

lhenze:
This PR addresses this bug: https://github.com/inveniosoftware/invenio-app-rdm/issues/1144

- Standardizes classname (with rest of site)
- hardcodes width and height to meet ORCID's branding guidelines
- Alt tag for a11y
- Adds span tag wrapper for ease of styling

### DIFF
--- a/src/lib/components/Creatibutors/CreatibutorsFieldItem.js
+++ b/src/lib/components/Creatibutors/CreatibutorsFieldItem.js
@@ -110,7 +110,7 @@ export const CreatibutorsFieldItem = ({
               {_get(initialCreatibutor, 'person_or_org.identifiers', []).some(
                 (identifier) => identifier.scheme === 'orcid'
               ) && (
-                <img className="inline-orcid" src="/static/images/orcid.svg" />
+                <img className="inline-id-icon" src="/static/images/orcid.svg" />
               )}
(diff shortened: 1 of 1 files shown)


src/lib/components/Creatibutors/CreatibutorsFieldItem.js
```diff
@@ -107,12 +108,14 @@ export const CreatibutorsFieldItem = ({
         <Ref innerRef={preview}>
           <List.Content>
             <List.Description>
+              <span class="creatibutor">
               {_get(initialCreatibutor, 'person_or_org.identifiers', []).some(
                 (identifier) => identifier.scheme === 'orcid'
               ) && (
-                <img className="inline-orcid" src="/static/images/orcid.svg" />
+                <img alt="ORCID logo" className="inline-id-icon" src="/static/images/orcid.svg" width="16" height="16" />
```
fenekku, line 115: We can add the `width` and `height` to the `inline-id-icon` css definition in a separate PR since ROR allows 16px height icon. This way, the icon should be consistent across display like https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/macros/creatibutors.html#L17 . We should also change the `img alt` there too.

fenekku (approved):

PR #406, opened 2022-01-13
1146 arrayfield item deletion

fenekku:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1146
- ArrayFields: fix deletion of item in an ArrayField
- release: v0.17.4
- requires released: https://github.com/inveniosoftware/react-invenio-forms/pull/99

### DIFF
--- a/src/lib/components/FundingField.js
+++ b/src/lib/components/FundingField.js
@@ -1,6 +1,6 @@
 // This file is part of React-Invenio-Deposit
 // Copyright (C) 2020 CERN.
-// Copyright (C) 2020 Northwestern University.
+// Copyright (C) 2020-2022 Northwestern University.
 // Copyright (C) 2021 Graz University of Technology.
 //
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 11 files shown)


src/lib/components/FundingField.js
```diff
@@ -51,84 +51,88 @@ export class FundingField extends Component {
...
-                icon
-                onClick={() => arrayHelpers.remove(indexPath)}
-              >
-                <Icon name="close" size="large" />
-              </Button>
-            </Form.Field>
-          </GroupField>
-        )}
+        {({ arrayHelpers, indexPath, form }) => {
+          const fieldPathPrefix = `${fieldPath}.${indexPath}`;
```
ntarocco, line 55: In the various components changed here and above, this is the only change right?

ntarocco (approved):
LGTM!

src/lib/components/FundingField.js
fenekku: That's right. The other parameters were not used, so I removed them and the indentation was adjusted. Nothing else.

PR #410, opened 2022-01-25
Creators auto-complete

### LINKED ISSUE #397: creatibutors modal: add support suggestion and autofill
- [x] Add a top level search field, 
Deposit__Creators_auto-completion_4

- [x] Display suggestions
Deposit__Creators_auto-completion_2

- [x] Autofill the form when one of the suggestions is chosen
Deposit__Creators_auto-completion_3
ld

- [x] Make modal collapsable
Deposit__Creators_auto-completion_1

slint:
Changes introduced from the creatibutors auto-complete feature branch. Some notes for the reviewer:

- We went with using Refs (which is horrible practice), since refactoring the underlying components to make state change handlers parametrized via props ended up being very messy/difficult, with unknown implications for other uses. If this is too hack-ish, we're up for refactoring, but we'll need some prior discussion/guidance for the specifics.

(Closes #397)

### DIFF
--- a/src/lib/components/Creatibutors/CreatibutorsModal.js
+++ b/src/lib/components/Creatibutors/CreatibutorsModal.js
@@ -15,6 +15,7 @@ import {
   TextField,
   ActionButton,
   RadioField,
+  RemoteSelectField,
 } from 'react-invenio-forms';
 import * as Yup from 'yup';
 import _get from 'lodash/get';
@@ -39,6 +40,8 @@ export class CreatibutorsModal extends Component {
       action: null,
     };
     this.inputRef = createRef();
+    this.identifiersRef = createRef();
+    this.affiliationsRef = createRef();
   }
 
   CreatorSchema = Yup.object({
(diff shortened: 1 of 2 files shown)


src/lib/components/Creatibutors/CreatibutorsModal.js
```diff
@@ -183,6 +187,124 @@ export class CreatibutorsModal extends Component {
...
+        [givenNameFieldPath]: selectedSuggestions[0].extra.given_name,
+        [familyNameFieldPath]: selectedSuggestions[0].extra.family_name,
+        [identifiersFieldPath]: identifiers,
+        [affiliationsFieldPath]: affiliations,
+      };
+      Object.entries(chosen).forEach(([path, value]) => {
+        formikProps.form.setFieldValue(path, value);
+      });
+      // Update identifiers render
+      this.identifiersRef.current.setState({
```
ntarocco, line 289: From a quick look (but I am probably wrong), in general this is a sign that the state should be handled by the upper component.
If you remove the state `selectedOptions` from `CreatibutorsIdentifiers`, and you manage it in this cmp, you should have here:
```
this.setState({selectedIdentifer: ....})
```
in the render method, you pass it as a prop, so that `CreatibutorsIdentifiers` will re-render with this new `selectedIdentifier`. You will also handle the callbacks and formik set in the upperCmp: the inner one will become very dummy.

If the logic in this upper becomes very messy, you can always introduce a middle cmp, between the 2, where the middle one encapsulates/manages the state of the inner one and the upper one is a bit lighter...
I had to do something similar in the PIDs cmp.

Happy to chat IRL :) I hope it helps (even if I am very positive that you have already thought about this).

slint:
Squash-merged manually in https://github.com/inveniosoftware/react-invenio-deposit/commit/04eba96448077d5242efa4ac947dba9b9d40bf8c (but GitHub doesn't detect that...)

PR #418, opened 2022-02-15
Community: add community header

jrcastro2:
Screenshots in https://github.com/inveniosoftware/invenio-app-rdm/pull/1260


src/lib/components/Communities/CommunityHeader.js
```diff
@@ -0,0 +1,138 @@
...
+CommunityHeaderInfo.defaultProps = {
+  community: undefined,
+};
+
+class EmptyCommunity extends React.Component {
+  render() {
+    const { error } = this.props;
+    return (
+      <Container className="community-header empty">
+        <span className={error && 'error-message'}>
```
kpsherva, line 69: why not to use the `<Message>` component instead?

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+class CommunityHeaderInfo extends React.Component {
+  render() {
+    const { community, imagePlaceholderLink } = this.props;
+    const logoLink = community ? community?.links?.logo : imagePlaceholderLink;
+    return (
+      <Container className="community-header">
+        <Image size="mini" src={logoLink} /> {community?.metadata?.title}
```
kpsherva, line 48: what about missing logos ? I think @zzacharo had a solution for this

src/lib/state/actions/communities.js
```diff
@@ -0,0 +1,52 @@
...
+const axiosWithconfig = axios.create(apiConfig);
+
+export const getCommunity = (communityId) => {
+  return async (dispatch, getState, config) => {
+    dispatch({
+      type: FETCH_COMMUNITY_LOADING,
+    });
+    try {
+      const response = await axiosWithconfig.get(
+        `/api/communities/${communityId}`
```
kpsherva, line 29: hardcoded url

src/lib/state/reducers/communities.js
```diff
@@ -0,0 +1,40 @@
...
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import {
+  FETCH_COMMUNITY_ERROR,
+  FETCH_COMMUNITY_LOADING,
+  FETCH_COMMUNITY_SUCCESS,
+  SET_COMMUNITY,
+} from '../types';
+
+export default (state = {}, action) => {
```
kpsherva, line 14: please name the reducer, it will be much easier to search for it in the code

src/lib/state/reducers/communities.js
```diff
@@ -0,0 +1,40 @@
...
+  switch (action.type) {
+    case FETCH_COMMUNITY_LOADING:
+      return {
+        ...state,
+        error: {},
+        loading: true,
+      };
+    case SET_COMMUNITY:
```
kpsherva, line 22: empty case?

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  render() {
+    const { loading, community, imagePlaceholderLink, error } = this.props;
+
+    return (
+      <Container className="deposits-community-header" fluid>
+        {loading ? (
+          <CommunityHeaderLoading />
```
kpsherva, line 110: it would be more readable if we make `CommunityHeaderLoading` as HOC, see implementation in ILS of `Loader` component. It will help you to avoid at least one level of this conditional statement

src/lib/components/Communities/CommunityHeader.js
zzacharo: I implemented a component here https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/components/Image.js

src/lib/components/Communities/CommunityHeader.js
jrcastro2: I personally like it less than just red text:
With message:
image

Just red text:
image

src/lib/components/Communities/CommunityHeader.js
jrcastro2: Hmm yes, I guess we can create this cmp in react-invenio-forms?, since it will be used widely (I guess)

src/lib/state/reducers/communities.js
jrcastro2: If it's empty it will return the same as the one below `FETCH_COMMUNITY_SUCCESS`

src/lib/state/reducers/communities.js
jrcastro2: Good point! Will rename also the others :+1:

src/lib/state/actions/communities.js
jrcastro2: Not sure, what's the best practice to avoid it ... any suggestions?

src/lib/components/Communities/CommunityHeader.js
jrcastro2: Resolved no error handling needed here (at the moment)

src/lib/state/actions/communities.js
jrcastro2: No need anymore

src/lib/components/Communities/CommunityHeader.js
jrcastro2: not needed anymore

zzacharo (approved):

src/lib/components/Communities/CommunityHeader.js
kpsherva: is the issue created?

ntarocco (approved):

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export class CommunityHeaderComponent extends React.Component {
+  componentDidMount() {
+    const { community, setCommunity } = this.props;
+    setCommunity(community);
+  }
+
+  render() {
+    const { communityRedux, imagePlaceholderLink } = this.props;
```
kpsherva, line 67: sorry naming police. I would say component should not know about redux even from names, why do we have `community` and `communityRedux`? could it be one prop ? if there is this separation maybe a parent should pass community to this component
```suggestion
    const { communityRedux, imagePlaceholderLink } = this.props;
```

jrcastro2: The `community` is the initial value and the `communityRedux` is the value stored in the client, I kept both as we will need to compare the values to check if they changed and do different actions in consequence (if I am not mistaken). 
We could probably rename it to `initialCommunity` and might be needed to be stored in redux too.

kpsherva: why do we need to compare the initial community and the one in redux ? 
I am probably missing some context of how this app works, but why not to store the initial community in redux when the whole app is initalized first and then manage the state and changes from redux?

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -0,0 +1,94 @@
...
```
ntarocco, line 69: Next PR: please change name. We don't want to use names that are related to technical implementation, but want to use name related to the domain language. `userSelectedCommunity`, `temporarlySelectedCommunity`, etc.., for example, might be good choices.

ntarocco (approved):

src/lib/components/Communities/CommunityHeader.js
kpsherva: the reason of this implementation was explained in a call with @zzacharo

PR #423, opened 2022-02-21
community header: Fix missing import

jrcastro2:
* Replaces semantic ui react Image with
  Image of react-invenio-forms

### DIFF
--- a/src/lib/components/Communities/CommunityHeader.js
+++ b/src/lib/components/Communities/CommunityHeader.js
@@ -7,15 +7,17 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import React from 'react';
-import { Container, Image, Button } from 'semantic-ui-react';
+import { Container, Button } from 'semantic-ui-react';
 import PropTypes from 'prop-types';
+import { i18next } from '@translations/i18next';
+import { Image } from 'react-invenio-forms';
 
 class CommunityHeaderButton extends React.Component {
   render() {
     return (
       <Button
         size="mini"
-        className="ml-5"
+        className="community-header-button ml-5"
         color="blue"
         onClick={() => console.log('TODO: Open community modal')}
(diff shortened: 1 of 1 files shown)


src/lib/components/Communities/CommunityHeader.js
```diff
@@ -32,8 +34,11 @@ class CommunityHeaderInfo extends React.Component {
     const logoLink = community ? community?.links?.logo : imagePlaceholderLink;
     return (
       <Container className="community-header">
-        <Image size="mini" src={logoLink} /> {community?.metadata?.title}
-        <CommunityHeaderButton />
+        <Image className="community-logo-header" src={logoLink} />
```
zzacharo, line 37: you miss the fallback url e.g point to `/static/images/sqaure-placeholder.png`?

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -32,8 +34,11 @@ class CommunityHeaderInfo extends React.Component {
...
+        <div className='mt-10'>
+          {community?.metadata?.title}
```
zzacharo, line 39: Creating a community I think requires a title. Can you check please and if it's the case then you should use `community.metadata.title` :)

src/lib/components/Communities/CommunityHeader.js
jrcastro2: Right, I was actually checking for the fallback above with:
```
const logoLink = community ? community?.links?.logo : imagePlaceholderLink;
```
But is not anymore needed.

src/lib/components/Communities/CommunityHeader.js
jrcastro2: Right, it's a leftover. Thanks!

ntarocco (approved):

PR #427, opened 2022-02-22
publish: disable btn on form changes

ntarocco:
* the `dirty` prop of Formik stores if the form was changed.
* for text field, the setFieldValue is called only onBlur, when the
  cursor leaves the field.
* if the user edit the text field, without leaving it, the `dirty` prop
  will not change value.
* partially fixes https://github.com/inveniosoftware/invenio-app-rdm/issues/794

### DIFF
--- a/src/lib/components/PublishButton.js
+++ b/src/lib/components/PublishButton.js
@@ -5,17 +5,16 @@
 // React-Invenio-Deposit is free software; you can redistribute it and/or modify it
 // under the terms of the MIT License; see LICENSE file for more details.
 
+import { i18next } from '@translations/i18next';
 import _get from 'lodash/get';
 import _isEmpty from 'lodash/isEmpty';
 import React, { Component } from 'react';
-import { connect } from 'react-redux';
-import { Icon, Button, Modal } from 'semantic-ui-react';
 import { ActionButton } from 'react-invenio-forms';
-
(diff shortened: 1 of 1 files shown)


ntarocco:
Corner case:
1. the user edits/changes a text field
2. the user leaves the field, without putting the focus anywhere else
3. the user clicks the `Publish` button
4. the `Publish` btn becomes disabled

https://user-images.githubusercontent.com/2089455/155186771-16dfbfd0-7421-4bdb-a47e-8a7729bd81d0.mp4

This is because, when the user clicks on the `Publish` btn:
1. the `onBlur` of the text field triggers
2. the `dirty` of the form becomes `True`
3. the `Publish` button is disabled **before** the click event

Possible workarounds:
1. we refactor the form that the `Publish` btn triggers a `Save` first. This would fix everything?
2. we add a small tooltip to the disabled `Publish` btn explaining why (fix errors/save first/no files uploaded yet), see how [here](https://github.com/Semantic-Org/Semantic-UI/issues/4296)
3. we add a label close to the btn saying `Unsaved changes` when there are unsaved changes, at least is explains why the `Publish` btn is disabled.

ntarocco:
After discussions, this is not a great solution.
The ideal solution, which will be implemented later on is:
1. enable form auto-save (optional)
2. automatically save when publishing

PR #431, opened 2022-02-23
communityHeader: add temporal solution for non-resolved community

jrcastro2:


zzacharo (approved):

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -31,15 +31,22 @@ class CommunityHeaderButton extends React.Component {
...
+      : imagePlaceholderLink;
+    const communityTitle = community?.metadata?.title
+      ? community.metadata.title
+      : community;
     return (
       <Container className="community-header">
         <Image
           className="community-logo-header"
-          src={community.links.logo}
+          src={communityLogo}
```
NRodriguezcuellar, line 45: Do you think this could also be

```js
src={community?.links?.logo}
```

because it should use the fallbackSrc if it is undefined right?

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -31,15 +31,22 @@ class CommunityHeaderButton extends React.Component {
...
           fallbackSrc={imagePlaceholderLink}
         />
         <div className="community-header-info">
-          {community.metadata.title}
+          {communityTitle}
```
NRodriguezcuellar, line 49: Community is an object right? i think it will error if it is passed instead of the `community.metadata.title` string

src/lib/components/Communities/CommunityHeader.js
jrcastro2: That's what I thought but doesn't seem the case, if it's undefined it will display a broken img.

src/lib/components/Communities/CommunityHeader.js
jrcastro2: Depends, it should always be an object but in some cases we can not yet resolve it and we only have the UUID value so we pass this as the `community` value. that's why we check for an object or otherwise we assume it's a string. This is a temporary solution, should be removed as soon as community is resolved in the backend, hopefully in next sprint :)

NRodriguezcuellar: ah ok! That is needed context, thanks.

NRodriguezcuellar (approved):

ntarocco (approved):

PR #446, opened 2022-03-09
creatibutors: make contributors family name required, fix double clic…

### LINKED ISSUE #439: Creatibutors modal: Name required issues
- The name/family-name for organization and person should also be marked as required for contributors (not only creators)

__Related bug:__
- When clicking save without entering a name (currently only for creators), a required-error shows up. When entering the name and clicking save, the error shows up again, but when clicking save again, the creator is saved. 

__Screenshots of the current flow__

1. Clicking save without entering name
Screenshot 2022-03-02 at 17 27 24
2. Entering name
Screenshot 2022-03-02 at 17 31 10
3. Clicking save
Screenshot 2022-03-02 at 17 27 41
4. Clicking save again will save the creator with no issues.

jennur:
Closes #439

- Added required to family/org name for both creator and contributor.
- Changed validation to validate onChange, but added `attemptedSubmit` state to not trigger the warnings unless the form was attempted submitted. This solves the wrongly displayed warning of no input in the name field when clicking save after first attempting to submit with no name value. Most likely because the value was not updated until the validation was run on submit. (Let me know if there is a better way to do this!)
- Fixed some small errors due to class vs className usage and wrong prop-value (boolean in className).

### DIFF
--- a/src/lib/components/Creatibutors/CreatibutorsModal.js
+++ b/src/lib/components/Creatibutors/CreatibutorsModal.js
@@ -49,6 +49,7 @@ export class CreatibutorsModal extends Component {
         props.autocompleteNames !== NamesAutocompleteOptions.SEARCH_ONLY ||
         !_isEmpty(props.initialCreatibutor)
       ),
+      attemptedSumbit: false
     };
     this.inputRef = createRef();
     this.identifiersRef = createRef();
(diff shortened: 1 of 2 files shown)


lnielsen:
@ntarocco I've added this one to your sprint board for review.

NRodriguezcuellar (commented):
Looks good to me, tried to think about other possible implementations because it seems weird to me that the validation does not happen after the values have been updated (before your fix). Maybe someone else has an idea?

Pineirin (approved):
LGTM!

ntarocco:
I would like to spend a bit of time to test other solutions, if you don't mind :)

src/lib/components/Creatibutors/CreatibutorsModal.js
```diff
@@ -335,7 +337,7 @@ export class CreatibutorsModal extends Component {
         onSubmit={this.onSubmit}
         enableReinitialize
         validationSchema={this.CreatorSchema}
-        validateOnChange={false}
+        validateOnChange={true}
```
ntarocco, line 340: I have done a few tests and only with this change (without the `attemptedSumbit` fix) it looks like working.
Would you have some time to check this together?

jennur: Yes, it works that way. The `attemptedSubmit` is just there to prevent the error messages to pop up when moving on to the next field :)

ntarocco:
Closed in favor of: https://github.com/inveniosoftware/react-invenio-forms/pull/114

PR #506, opened 2022-05-30
dropdowns: fix missalignment

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1637

image

image

### DIFF
--- a/src/lib/components/Funding/FunderDropdown.js
+++ b/src/lib/components/Funding/FunderDropdown.js
@@ -11,62 +11,68 @@ import { Dropdown } from 'semantic-ui-react';
 import { withState } from 'react-searchkit';
 import { i18next } from '@translations/i18next';
 
-export const FunderDropdown = withState(({ currentResultsState: awardsList, updateQueryState: updateQueryState, currentQueryState: currentQueryState }) => {
+export const FunderDropdown = withState(
+  ({
+    currentResultsState: awardsList,
+    updateQueryState: updateQueryState,
+    currentQueryState: currentQueryState,
+  }) => {
     const [fundersFromFacets] = useFundersFromFacets(awardsList);
-  
+
     /**
      * Trigger on funder selection.
      * Updated the query state to filter by the selected funder.
-     * 
-     * @param {*} event 
-     * @param {*} data 
+     *
+     * @param {*} event
+     * @param {*} data
      */
     function onFunderSelect(event, data) {
       let newFilter = [];
-  
-      if (data && data.value !== "") {
-        newFilter = ['funders', data.value]
+
+      if (data && data.value !== '') {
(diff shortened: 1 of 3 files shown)


src/lib/components/Funding/FunderDropdown.js
```diff
@@ -11,69 +11,74 @@ import { Dropdown } from 'semantic-ui-react';
...
     }
-  
+
     return (
       <Dropdown
         placeholder={i18next.t('Funder')}
         search
         selection
         clearable
-        scrolling
```
kpsherva, line 76: how does it display with removed scrolling and large number of results ?

jrcastro2: It still displays a scroll (I am testing with mocked data)

image

with `scrolling` props the dropdown was not so big, but if we add it, it will cause the missalignment, another option would be to set a max-height using css instead of the `scrolling` props

kpsherva: I think it is better to find the cause of the misalignment and fix the css while using scrolling, rather to find workarounds on how not to use scrolling :) 
think about this in terms of someone else implementing another new dropdown, and reintroducing the misalignment yet again. Isn't it better to fix it once for all ?

jrcastro2: Well, I found the cause of the issue but I am not sure it's us ... Actually in a first look I would say it's a semantic ui issue and fixing it will probably imply to change the css when the `scrolling` prop is applied. I am saying that the error is on their side because when applying the scrolling, we also apply the following `min-width: 100% !important;` (see [code here](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.less#L1081)) and this overrides the folldowing rule ` min-width: calc(100% + 2px );` (see [code here](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.less#L419)). We need that extra 2px to avoid the missalignment. That said if we agree, I can override the `.ui.selection.dropdown .menu`  to have the same correct width, but again, we are introducing custom css, that's what I was trying to avoid by removing the `scrolling` prop.

kpsherva: it is totally OK to introduce custom CSS if it is fixing a global issue - that's why we have a centrally managed theming. The problem with custom CSS appears when it is duplicated across different components or created to "reinvent the wheel" while the main theme already provides the needed styling

jrcastro2: Added back the scrolling and fixed the rule in invenio-theme, [here the PR.](https://github.com/inveniosoftware/invenio-theme/pull/290)

kpsherva (approved):

PR #511, opened 2022-06-01
PIDField: fix display of initial errors

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1148

image

### DIFF
--- a/src/lib/components/Identifiers/PIDField.js
+++ b/src/lib/components/Identifiers/PIDField.js
@@ -22,6 +22,13 @@ import { DISCARD_PID_STARTED, RESERVE_PID_STARTED } from '../../state/types';
 const PROVIDER_EXTERNAL = 'external';
 const UPDATE_PID_DEBOUNCE_MS = 200;
 
+const getFieldErrors = (form, fieldPath) => {
+  return (
+    getIn(form.errors, fieldPath, null) ||
+    getIn(form.initialErrors, fieldPath, null)
+  );
+};
+
 /**
  * Button component to reserve a PID.
  */
@@ -301,25 +308,16 @@ class UnmanagedIdentifierCmp extends Component {
   render() {
     const { localIdentifier } = this.state;
     const { form, fieldPath, helpText, pidPlaceholder } = this.props;
-
+    const fieldError = getFieldErrors(form, fieldPath);
     return (
       <>
-        <Form.Field
-          width={8}
-          error={
-            getIn(form.errors, fieldPath, null) ||
-            getIn(form.errors, 'pids', null)
-          }
-        >
+        <Form.Field width={8} error={fieldError}>
           <Form.Input
             onChange={(e, { value }) => this.onChange(value)}
             value={localIdentifier}
             placeholder={pidPlaceholder}
             width={16}
-            error={
-              getIn(form.errors, fieldPath, null) ||
-              getIn(form.errors, 'pids', null)
-            }
+            error={fieldError}
           />
         </Form.Field>
         {helpText && <label className="helptext">{helpText}</label>}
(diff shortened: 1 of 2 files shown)


kpsherva (commented):
I think I will need a call to understand how this works :)

jrcastro2:
> I think I will need a call to understand how this works :)

I can try to explain it here so everyone can check it but feel free to reach out if not clear enough.

We also need to check for the `initialErrors` mainly because we have added the formik option [`enableReinitialize`](https://formik.org/docs/api/withFormik#enablereinitialize-boolean) that apparently resets the form when the props change (when submitting as example). Meaning that every time the props change the form resets and the `Errors` prop is cleared. In order to still keep the errors we store them in `InitialErrrors` (see ["workaround" here](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositBootstrap.js#L124)). So when we display the errors, first we need to check for the `Errors` prop of Formik and if empty we also need to check for the `InitialErrrors` prop as the errors might be cleared due to `enableReinitialize`.

src/lib/components/Identifiers/PIDField.js
```diff
@@ -308,7 +308,8 @@ class UnmanagedIdentifierCmp extends Component {
           width={8}
           error={
             getIn(form.errors, fieldPath, null) ||
```
NRodriguezcuellar, line 310: minor code style thing: 
```js
const fieldError = getIn(form.errors, fieldPath, null) || getIn(form.errors, 'pids', null) ||
            getIn(form.initialErrors, fieldPath, null)
```
which can be re-used for here and down below

NRodriguezcuellar (approved):

Pineirin (approved):

kpsherva (approved):

src/lib/components/Identifiers/PIDField.js
```diff
@@ -301,25 +301,19 @@ class UnmanagedIdentifierCmp extends Component {
   render() {
     const { localIdentifier } = this.state;
     const { form, fieldPath, helpText, pidPlaceholder } = this.props;
-
+    const fieldError =
+      getIn(form.errors, fieldPath, null) ||
```
zzacharo, line 305: nit: Maybe we extract this into a function to avoid duplication?

src/lib/components/Identifiers/PIDField.js
```diff
@@ -301,25 +301,19 @@ class UnmanagedIdentifierCmp extends Component {
...
+      getIn(form.errors, 'pids', null) ||
```
zzacharo, line 306: Isn't fieldPath set to `pids`?

jrcastro2: The fieldpath is set to `fieldPath={'pids.${pid.scheme}'}` ([see code here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/js/invenio_app_rdm/deposit/RDMDepositForm.js#L147)). I think we could remove the check for the hardcoded `pids` but not sure if this will cause other bugs as we are only using DOIs in the `pids` field. Maybe it's needed if other type of PIDs are registered? If you give the green light I will remove it

zzacharo: I tried in our QA environment and I only see errors returned for the fieldPath rather than `pids` field.... I would say is safe to remove it now

jrcastro2: Ok, thanks! Done

zzacharo (approved):

src/lib/components/Identifiers/PIDField.js
```diff
@@ -22,6 +22,13 @@ import { DISCARD_PID_STARTED, RESERVE_PID_STARTED } from '../../state/types';
 const PROVIDER_EXTERNAL = 'external';
 const UPDATE_PID_DEBOUNCE_MS = 200;
 
+const getFieldErrors = (form, fieldPath) => {
+  return (
+    getIn(form.errors, fieldPath, null) ||
+    getIn(form.initialErrors, fieldPath, null)
```
zzacharo, line 28: I think you can remove the label also from here https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/FormFeedback.js#L51

kpsherva (approved):

PR #517, opened 2022-06-03
form feedback: show error response from backend

NRodriguezcuellar:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1692

error message from backend
Screenshot from 2022-06-03 09-28-43

usual validation messages (to show they still work)
Screenshot from 2022-06-03 09-42-34
Screenshot from 2022-06-03 09-42-26

### DIFF
--- a/src/lib/DepositApiClient.js
+++ b/src/lib/DepositApiClient.js
@@ -97,12 +97,8 @@ export class RDMDepositApiClient extends DepositApiClient {
       );
       return new DepositApiClientResponse(data, errors);
     } catch (error) {
-      const data = this.recordSerializer.deserialize(error.response.data || {});
-      const errors = this.recordSerializer.deserializeErrors(
-        error.response.data.errors || []
-      );
-      const wrapped = new DepositApiClientResponse(data, errors);
-      throw new Error(wrapped);
+      const errorData = error.response.data;
+      throw new DepositApiClientResponse({}, errorData);
     }
   }
 
@@ -284,6 +280,7 @@ export class DepositFileApiClient {
       throw new Error('Abstract');
     }
   }
+
   isCancelled(error) {
     return axios.isCancel(error);
   }
--- a/src/lib/components/FormFeedback.js
+++ b/src/lib/components/FormFeedback.js
@@ -233,10 +233,12 @@ class DisconnectedFormFeedback extends Component {
 
     return labelledErrorMessages;
   }
+
   render() {
     const actionState = this.props.actionState;
 
     const errors = this.props.errors || {};
+
     const { feedback, message } = _get(ACTIONS, actionState, {
       feedback: undefined,
       message: undefined,
@@ -256,6 +258,9 @@ class DisconnectedFormFeedback extends Component {
       )
     );
 
+    // errors not related to validation, following a different format {status:.., message:..}
+    const backendErrorMessage = errors.message;
+
     return (
       <Message
         visible
@@ -266,7 +271,7 @@ class DisconnectedFormFeedback extends Component {
       >
         <Grid container>
           <Grid.Column width={15} textAlign="left">
-            <strong>{message}</strong>
+            <strong>{backendErrorMessage || message}</strong>
             {listErrors.length > 0 && <Message.List>{listErrors}</Message.List>}
           </Grid.Column>
         </Grid>


src/lib/DepositApiClient.js
```diff
@@ -97,12 +97,20 @@ export class RDMDepositApiClient extends DepositApiClient {
       );
       return new DepositApiClientResponse(data, errors);
     } catch (error) {
-      const data = this.recordSerializer.deserialize(error.response.data || {});
+      const responseData = error.response.data;
+      const data = this.recordSerializer.deserialize(responseData || {});
+
+      const noFieldErrors = !('errors' in responseData);
```
alejandromumo, line 103: minor: `prop in object` operator checks for `prop` in an object **and its prototype**. It is safer to look using `object.hasOwnProperty('prop')` if inheritance from prototype is not desired.

src/lib/DepositApiClient.js
```diff
@@ -97,12 +97,20 @@ export class RDMDepositApiClient extends DepositApiClient {
...
+
+      if (noFieldErrors) {
+        throw new DepositApiClientResponse(data, responseData);
+      }
+
```
alejandromumo, line 108: suggestion: this method could be rewritten in 12 lines with a simplified logic:

```javascript
async _createResponse(axiosRequest) {
    let responseData = {}
    try {
        const response = await axiosRequest();
        responseData = response.data;
    } catch (error) {
        responseData = error.response.data;
    }
    const data = this.recordSerializer.deserialize(responseData || {});
    const errors = this.recordSerializer.deserializeErrors(
        response.data.errors || []
    );
    return new DepositApiClientResponse(data, errors);
}
```

NRodriguezcuellar: i tried it but then i realized that it either needs to return or throw the final value otherwise it will not work correctly

alejandromumo: In that case adding a boolean flag would be enough.

```javascript
let isThrowing = false;

try {
    ...
}
catch{
  ... 
  isThrowing = true;
}

if (isThrowing) {
    throw new ...
}

return ...
```

Note that this is a suggestion on code readability, it is not an optimization of any sorts so it is completely a suggestion that can be ignored if you want :)

Pineirin (approved):
Looks right to me.

src/lib/DepositApiClient.js
NRodriguezcuellar: thank you!

src/lib/DepositApiClient.js
```diff
@@ -89,21 +89,33 @@ export class RDMDepositApiClient extends DepositApiClient {
...
-      );
-      const wrapped = new DepositApiClientResponse(data, errors);
-      throw new Error(wrapped);
+      shouldThrowError = true;
+      responseData = error.response.data;
     }
+
+    const data = this.recordSerializer.deserialize(responseData || {});
+
+    const noFormFieldErrors =
```
ntarocco, line 105: Very minor: best practice is to go for "positive" vars, it is easier to understand while reading it.
```suggestion
    const formHasErrors = ...
```

src/lib/DepositApiClient.js
```diff
@@ -89,21 +89,33 @@ export class RDMDepositApiClient extends DepositApiClient {
...
+      !responseData.hasOwnProperty('errors') && shouldThrowError;
+
+    const errors = noFormFieldErrors
```
ntarocco, line 108: I am confused by this. If no form errors, the errors is the `responseData`?

General feeling: I have trouble understanding the code by quickly reading it, with multiple`if`s around? Wasn't it more readable before (even if there was some little duplication)?

NRodriguezcuellar: i agree that it was a bit too simplified, but i would argue that it is more descriptive than before now.

alejandromumo: I agree with @ntarocco , the initial suggestion was meant to improve readability. But I feel we traded duplication for conditional logic, especially ternary operators, which make the code a bit harder to read right now.

Most of the conditional logic might not be needed, since JS already does it for us when you write code like:

```javascript
const errors = this.recordSerializer.deserializeErrors(responseData.errors || [])
```

I'd argue only one `if` statement is needed for the whole function, but we can discuss it @NRodriguezcuellar.

Also, feel free to discard my suggestions if they do not improve readability at all.

src/lib/DepositApiClient.js
```diff
@@ -97,12 +98,14 @@ export class RDMDepositApiClient extends DepositApiClient {
...
       return new DepositApiClientResponse(data, errors);
     } catch (error) {
-      const data = this.recordSerializer.deserialize(error.response.data || {});
-      const errors = this.recordSerializer.deserializeErrors(
-        error.response.data.errors || []
-      );
-      const wrapped = new DepositApiClientResponse(data, errors);
-      throw new Error(wrapped);
+      const errorData = error.response.data;
+      const formHasFieldErrors = errorData.hasOwnProperty('errors');
```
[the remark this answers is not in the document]
NRodriguezcuellar: you're right!

NRodriguezcuellar:
comment

src/lib/DepositApiClient.js
```diff
@@ -6,6 +6,7 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import axios from 'axios';
+import _isEmpty from 'lodash/isEmpty';
```
zzacharo, line 9: this can be removed I think now, no?

src/lib/DepositApiClient.js
```diff
@@ -97,12 +98,9 @@ export class RDMDepositApiClient extends DepositApiClient {
...
       return new DepositApiClientResponse(data, errors);
     } catch (error) {
-      const data = this.recordSerializer.deserialize(error.response.data || {});
-      const errors = this.recordSerializer.deserializeErrors(
-        error.response.data.errors || []
-      );
-      const wrapped = new DepositApiClientResponse(data, errors);
-      throw new Error(wrapped);
+      const errorData = error.response.data;
+
```
zzacharo, line 102: remove the empty line?

zzacharo (approved):

kpsherva (approved):

PR #527, opened 2022-06-13
communities: fix logos having inconsistent heights

NRodriguezcuellar:
partly closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1704
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1681
needs https://github.com/inveniosoftware/invenio-app-rdm/pull/1710

Screenshot from 2022-06-13 14-31-36
Screenshot from 2022-06-13 14-31-30
Screenshot from 2022-06-13 14-31-23

Screenshot from 2022-06-13 14-11-42

### DIFF
--- a/src/lib/components/Communities/CommunityHeader.js
+++ b/src/lib/components/Communities/CommunityHeader.js
@@ -34,13 +34,13 @@ class CommunityHeaderComponent extends Component {
               <>
                 <div className="page-subheader-element">
                   <Image
-                    size="mini"
-                    className="community-logo-header"
+                    size="tiny"
+                    className="community-header-logo"
                     src={community.links?.logo || imagePlaceholderLink} // logo is undefined when new draft and no selection
                     fallbackSrc={imagePlaceholderLink}
                   />
                 </div>
-                <div className="page-subheader-element">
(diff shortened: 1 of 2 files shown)


Pineirin (approved):

src/lib/components/Communities/CommunityHeader.js
```diff
@@ -34,13 +34,13 @@ class CommunityHeaderComponent extends Component {
               <>
                 <div className="page-subheader-element">
                   <Image
-                    size="mini"
-                    className="community-logo-header"
+                    size="tiny"
```
kpsherva, line 37: the size of the image was changed many times and finally agreed on `mini`. How does it look like after this change when the image is not "tall" but with square ratio?

NRodriguezcuellar: I would say the first picture is pretty square, you didn't notice it? I changed it because before it could overflow, maybe it looked bigger but i seemed really small to me now.

jennur (approved):

kpsherva (approved):

PR #534, opened 2022-06-27
files: added string normalization for filenames comparison.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1744

The issue here boils down on how client and server deal with special characters for file names. E.g. the character á can be encoded as a single unicode character (á) or as a composite character (a ´). 

Therefore: 

- when comparing filenames between server and client, string normalisation is needed.
- when the server sends an event (e.g. `FILE_UPLOAD_CANCELLED`) to the client, the filename's string must be normalised as well. Otherwise, `state.entries` will index files by name and there will be a mismatch again and  two "files" are going to be added to `state.entries`.  

Note: this is a quick fix for this issue. Ideally we should be consistent with the encoding. In my opinion, the client should adapt to the server and not the other way around. Nonetheless, the file creation is handled by HTML5 / javascript and the created `File` object has the property `name` as `read-only` (see js [documentation](https://w3c.github.io/FileAPI/#dfn-name) for files). I tried changing the html page encoding but it did not fix the issue.

### DIFF
--- a/src/lib/state/reducers/files.js
+++ b/src/lib/state/reducers/files.js
@@ -32,15 +32,17 @@ const initialState = {};
 
 const fileReducer = (state = initialState, action) => {
   let newState;
+  // Filename needs to be normalised due to encoding differences between client and server.
+  const remoteFileName = action.payload?.filename?.normalize() ?? '';
   switch (action.type) {
     case FILE_UPLOAD_ADDED:
       return {
         ...state,
         entries: {
           ...state.entries,
-          [action.payload.filename]: {
+          [remoteFileName]: {
             progressPercentage: 0,
-            name: action.payload.filename,
+            name: remoteFileName,
             size: 0,
             status: UploadState.pending,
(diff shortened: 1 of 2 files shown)


kpsherva (approved):

src/lib/state/reducers/files.js
```diff
@@ -32,15 +32,17 @@ const initialState = {};
 
 const fileReducer = (state = initialState, action) => {
   let newState;
+  // Filename needs to be normalised due to encoding differences between client and server.
+  let remoteFileName = action.payload?.filename ? action.payload.filename.normalize() : '';
```
zzacharo, line 36: I think this can be reduced to 
```suggestion
  let remoteFileName = action.payload?.filename?.normalize() ?? '';
```

zzacharo (approved):

jrcastro2 (approved):

PR #547, opened 2022-07-05
Put semver.org link between parenthesis

avivace:
The help text for the version string is currently a bit strange to read:

```
Mostly relevant for software and dataset uploads. A semantic version string is preferred see [semver.org](https://semver.org/), but any version string is accepted.
```

This PR just puts the semver.org link between parenthesis so it's easier to follow the sense of the phrase.

### DIFF
--- a/src/lib/components/VersionField.js
+++ b/src/lib/components/VersionField.js
@@ -19,7 +19,7 @@ export class VersionField extends Component {
       <span>
         <Trans>
           Mostly relevant for software and dataset uploads. A semantic version
-          string is preferred see
+          string is preferred (see
           <a
             href="https://semver.org/"
             target="_blank"
@@ -28,7 +28,7 @@ export class VersionField extends Component {
             {' '}
             semver.org
           </a>
-          , but any version string is accepted.
+          ) but any version string is accepted.
         </Trans>
       </span>
     );


src/lib/components/VersionField.js
```diff
@@ -28,7 +28,7 @@ export class VersionField extends Component {
             {' '}
             semver.org
           </a>
-          , but any version string is accepted.
+          ) but any version string is accepted.
```
avivace, line 31:
```suggestion
          ) but any string is accepted.
```

Also, I suggest to remove "version string" to just use "string" here, because it better suggests that any string can be used as a version (while "version string" may suggest `x.y.z`). E.g. `revision-conference-80` is accepted in this field, as well as

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

without triggering any validation error

carlinmack:
Closing as this module is deprecated, you can contribute this pull request again in the new location of the module which is rdm-records https://github.com/inveniosoftware/invenio-rdm-records/tree/master/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src

Issue #549, opened 2022-07-07
records: fix record visibility when community is changed.


alejandromumo:
Currently record visibility is not changed when the community is selected in the deposit form. E.g. if a restricted community is selected, the record will still have "public" access even though the UI blocks the `Public` button.

After receiving an error from the backend, any change to the form will wipe the form as an unhandled exception is generated.

Additionally, the thrown exception does  not have an error handler as is returning an HTML containing the flask exception's trackeback. 

![unknown](https://user-images.githubusercontent.com/21204744/177769943-76db197f-0acb-4737-bc47-f6a1173e6b95.png)

![unknown2](https://user-images.githubusercontent.com/21204744/177769933-5864e469-0875-4377-ab84-1aba52f65231.png)

PR #550, opened 2022-07-07
records: fixed record access based on community visibility.

alejandromumo:
closes https://github.com/inveniosoftware/react-invenio-deposit/issues/549

### DIFF
--- a/src/lib/components/Access/ProtectionButtons.js
+++ b/src/lib/components/Access/ProtectionButtons.js
@@ -7,7 +7,7 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 import React, { Component } from 'react';
 import { Button } from 'semantic-ui-react';
-import { FastField } from 'formik';
+import { Field } from 'formik';
 import { i18next } from '@translations/i18next';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


alejandromumo:
Currently the community's visibility is not automatically tied to the record's access. This is particularlly troublesome when there is a change in the  draft's community.

Therefore, we decided that, for now, the best place to update the record access was in the deposit reducer after the event `SET_COMMUNITY` is dispatched.

src/lib/components/Access/ProtectionButtons.js
```diff
@@ -72,12 +72,10 @@ ProtectionButtonsComponent.defaultProps = {
...
 export class ProtectionButtons extends Component {
   render() {
-    const { fieldPath, active, disabled } = this.props;
+    const { fieldPath } = this.props;
 
-    const changed = !active || disabled ? { changed: true } : {};
     return (
-      <FastField
-        {...changed}
+      <Field
```
alejandromumo, line 78: This component was not re-rendering after the changes on the reducer.

It can be analysed later.

kpsherva (approved):

PR #551, opened 2022-07-07
eslint: add invenio eslint rules

NRodriguezcuellar:
partly closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1789


.eslintrc.yml
```diff
@@ -0,0 +1,24 @@
+extends:
+  - '@inveniosoftware/invenio'
+  - '@inveniosoftware/invenio/prettier'
+
+parser: '@babel/eslint-parser'
+
+rules:
+  # allow default prop to be set even if prop required
+  react/default-props-match-prop-types: 0
```
jrcastro2, line 9: The idea is to not add custom rules to each repo but to use the ones provided by the eslint repo to make sure that all the projects follow the same rules. That said if a rule is meant to be added should be in the [eslint repo](https://github.com/inveniosoftware/eslint-config-invenio). We already defined [here](https://github.com/inveniosoftware/eslint-config-invenio/blob/master/configs/main.yaml#L22) this rule to fail on purpose, why would we want to allow default props for required props?

package.json
```diff
@@ -97,7 +105,7 @@
     "react-invenio-forms": "^0.10.7",
     "react-overridable": "^0.0.3",
     "react-redux": "^7.2.0",
-    "react-scripts": "^3.4.0",
+    "react-scripts": "^5.0.1",
```
jrcastro2, line 108: Not sure about this changes, we would need to check breaking changes and maybe upgrade it in all other repos like [react-invenio-forms](https://github.com/inveniosoftware/react-invenio-forms/blob/master/package.json#L66). Are you sure you need all this changes in the `package.json`?

NRodriguezcuellar: These changes were made because otherwise eslint was failing to lint on my system/editor. Because they are dev dependencies my assumption is that they are only used for development or when building the library, that should make it safe to upgrade (we have some quite old dependencies). What do you think?

.eslintrc.yml
NRodriguezcuellar: That is a very good point, i did not realize it was set on purpose . I set it here because it seems to be a common pattern for some props and i was not sure about changing the actual code aside from formatting and some minor code refactoring (destructuring prop, adding prop types).

kpsherva: on another subject: do we have any specification on the line lenght ? It would be good to align with black(backend formatter)

jrcastro2: We use the default value which is 80 if I am not mistaken , but we can update it.

kpsherva: after the discussion, during the stand-up today, I think in the end a prop which has a default is not technically required, since it is not provided when the component is used, so it makes sense to have either default or required, not both

package.json
ntarocco: This should not be a problem, because they are `devDependencies`: when packaged and released on `npmjs.org`, the modules using this one will not install these dependencies.
What we need always to pay attention is `dependencies` and `peerDeps`.

ntarocco (commented):
Should be merged ASAP to avoid infinite amount of conflicts!
Ping @kpsherva @jrcastro2 @zzacharo

.eslintrc.yml
jrcastro2: No, not added. We quickly discussed during the stand-up and I personally agree with what @kpsherva explained in the previous comment, the idea is to keep the rule as it is, however this is open to discussion, might be faster IRL.

NRodriguezcuellar: indeed as @jrcastro2 said

src/lib/DepositBootstrap.js
```diff
@@ -158,10 +189,8 @@ const mapDispatchToProps = (dispatch) => ({
   previewAction: (values) => dispatch(preview(values)),
   deleteAction: (values, { isDiscardingVersion }) =>
     dispatch(delete_(values, { isDiscardingVersion })),
-  reservePIDAction: (values, { pidType }) =>
-    dispatch(reservePID(values, { pidType })),
-  discardPIDAction: (values, { pidType }) =>
-    dispatch(discardPID(values, { pidType })),
+  reservePIDAction: (values, { pidType }) => dispatch(reservePID(values, { pidType })),
+  discardPIDAction: (values, { pidType }) => dispatch(discardPID(values, { pidType })),
```
NRodriguezcuellar, line 193: these should all be required

src/lib/components/PublishButton/SubmitReviewModal.js
```diff
@@ -174,3 +154,7 @@ SubmitReviewModal.propTypes = {
   onSubmit: PropTypes.func.isRequired,
   initialReviewComment: PropTypes.string,
 };
+
+SubmitReviewModal.defaultProps = {
+  initialReviewComment: "",
```
kpsherva, line 159: will this result in displaying empty comments?

src/lib/DepositFormApp.js
```diff
@@ -112,4 +106,6 @@ DepositFormApp.defaultProps = {
   draftsService: null,
   filesService: null,
   recordSerializer: null,
+  files: undefined,
```
zzacharo, line 109: this can be described more accurately in the Proptypes e.g `Proptypes.shape(...)`

zzacharo: maybe is better to fallback to `undefined` always instead of null for all the missing props....Do we need to define value for optional properties actually? Can you test if removing them works? I mean the `null`/`undefined`...

src/lib/components/Access/EmbargoDateField.js
```diff
@@ -29,15 +29,17 @@ export function EmbargoDateField({
...
 
 EmbargoDateField.defaultProps = {
-  fieldPath: 'access.embargo.until',
-  label: i18next.t('Embargo until'),
-  labelIcon: 'calendar',
-  placeholder: i18next.t('YYYY-MM-DD'),
+  required: false,
+  labelIcon: "calendar",
+  placeholder: i18next.t("YYYY-MM-DD"),
+  fieldPath: "access.embargo.until",
```
zzacharo, line 43: if we upper pin invenio-app-rdm and we release these changes as `v0.20.x` then we should remove the `defaultProp` `fieldPath`...otherwise, you need here an ignore rule no? I would also maybe add a comment saying that this will be removed in the future or similar...

src/lib/components/AccessRightField.js
```diff
@@ -127,5 +120,8 @@ AccessRightField.propTypes = {
 };
 
 AccessRightField.defaultProps = {
-  fieldPath: 'access',
+  // eslint-disable-next-line react/default-props-match-prop-types
+  fieldPath: "access",
```
zzacharo, line 124: Check my comment above :)

src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
```diff
@@ -129,11 +128,3 @@ export class CommunitySelectionSearch extends Component {
     );
   }
 }
-
-CommunitySelectionSearch.propTypes = {
```
zzacharo, line 133: Why is this removed?

src/lib/components/DescriptionsField.js
```diff
@@ -35,17 +32,19 @@ export class DescriptionsField extends Component {
...
 
 DescriptionsField.defaultProps = {
-  fieldPath: 'metadata.description',
-  label: i18next.t('Description'),
-  labelIcon: 'pencil',
+  // eslint-disable-next-line react/default-props-match-prop-types
+  fieldPath: "metadata.description",
+  label: i18next.t("Description"),
+  labelIcon: "pencil",
   editorConfig: {},
```
zzacharo, line 48: maybe use `undefined` here too?

src/lib/components/Access/EmbargoDateField.js
zzacharo: Please take this into account for all the `eslint` ignore rules you added :)

NRodriguezcuellar: Yes, i looked over this one, i just made sure that every `fieldPath` prop is set to `isRequired`!

NRodriguezcuellar: maybe i should post a comment in this PR documenting why this decision (ignoring the fieldPaths defaultProps) was made and link it with every ignore?

src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
NRodriguezcuellar: because the components uses no props 😄

src/lib/DepositFormApp.js
NRodriguezcuellar: Yes it is a rule that every propType that is not `required` should have a defaultProp, as you see i usually default to undefined so that the behaviour does not change unexpectedly.

as for setting the shape i am not sure what that could be in this case 😅

src/lib/components/Access/EmbargoDateField.js
kpsherva: I would start an issue to clean it out in the 0.20 maybe

NRodriguezcuellar: issue made here https://github.com/inveniosoftware/react-invenio-deposit/issues/563

kpsherva (approved):

PR #565, opened 2022-08-01
global: Fixes wording of error-messages.

chriz-uniba:
Refs: https://github.com/inveniosoftware/invenio-app-rdm/issues/1707

### DIFF
--- a/src/lib/components/Funding/FundingModal.js
+++ b/src/lib/components/Funding/FundingModal.js
@@ -79,7 +79,7 @@ const CustomFundingSchema = Yup.object().shape({
         },
       }),
       url: Yup.string()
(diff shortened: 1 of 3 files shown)


src/lib/components/Funding/FundingModal.js
```diff
@@ -79,7 +79,7 @@ const CustomFundingSchema = Yup.object().shape({
         },
       }),
       url: Yup.string()
-        .url(i18next.t("Url must be a valid url."))
+        .url(i18next.t("Url must be valid."))
```
lnielsen, line 82:
```suggestion
        .url(i18next.t("URL must be valid."))
```

PR #583, opened 2022-10-04
deposit api: Config api headers 

philippgualdi:
The implementation allows you to configure the BASE_HEADERS for other modules using the DepositFormApp and the REST
api client.

### DIFF
--- a/src/lib/DepositApiClient.js
+++ b/src/lib/DepositApiClient.js
@@ -6,8 +6,8 @@
 // under the terms of the MIT License; see LICENSE file for more details.
 
 import axios from "axios";
+import _get from "lodash/get";
 
-const CancelToken = axios.CancelToken;
 const BASE_HEADERS = {
   "json": { "Content-Type": "application/json" },
   "vnd+json": {
@@ -16,13 +16,6 @@ const BASE_HEADERS = {
   },
   "octet-stream": { "Content-Type": "application/octet-stream" },
 };
-const apiConfig = {
-  withCredentials: true,
-  xsrfCookieName: "csrftoken",
-  xsrfHeaderName: "X-CSRFToken",
-  headers: BASE_HEADERS.json,
-};
-const axiosWithConfig = axios.create(apiConfig);
 
 /**
  * API client response.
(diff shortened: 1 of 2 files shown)


src/lib/DepositApiClient.js
```diff
@@ -111,8 +103,8 @@ export class RDMDepositApiClient extends DepositApiClient {
   async createDraft(draft) {
     const payload = this.recordSerializer.serialize(draft);
     return this._createResponse(() =>
-      axiosWithConfig.post(this.createDraftURL, payload, {
-        headers: BASE_HEADERS["vnd+json"],
+      this.axiosWithConfig.post(this.createDraftURL, payload, {
+        headers: this.apiHeaders["vnd+json"],
```
kpsherva, line 107: with BASE_HEADERS removed, we cannot assure `vnd+json` to be present in the apiHeaders object

philippgualdi: The BASE_HEADERS are set in the configuration in the app inveniosoftware/invenio-app-rdm#1917.

philippgualdi: The BASE_HEADERS will be set in the invenio_app_rdm and allows other modules to set their own headers.

kpsherva: I understand, however this solution relies on defining the headers in a repository which depends on react-invenio-deposit. As a result there is a "mutual dependency" created - module cannot be used without invenio-app-rdm. 
If the assumption of having `vnd+json` header in this.apiHeaders object is not fulfiled then headers will be evaluated as `undefined`

philippgualdi: I see your point and pushed a solution where no mutual dependency is created.

src/lib/DepositApiClient.js
```diff
@@ -36,10 +28,21 @@ export class DepositApiClientResponse {
...
   /* eslint-disable no-unused-vars */
-  constructor(createDraftURL, recordSerializer) {
+  constructor(apiHeaders, createDraftURL, recordSerializer) {
     if (this.constructor === DepositApiClient) {
       throw new Error("Abstract");
     }
+
+    this.apiHeaders = Object.assign({}, BASE_HEADERS, apiHeaders);
+
+    this.apiConfig = {
```
ntarocco, line 38: Suggestion: make the entire config overridable, something like this (pseudocode):
```
constructor(extraApiConfig, createDraftURL, recordSerializer) {
...
}

const extraHeaders = _get(extraApiConfig, "headers", {});
this.apiHeaders = Object.assign({}, BASE_HEADERS, extraHeaders);

this.apiConfig = {
    withCredentials: true,
    ...extraApiConfig,
}
```

src/lib/DepositFormApp.js
```diff
@@ -38,9 +38,11 @@ export class DepositFormApp extends Component {
           props.config.custom_fields.vocabularies
         );
 
+    const apiHeaders = props.config.apiHeaders ? props.config.apiHeaders : null;
+
     const apiClient = props.apiClient
       ? props.apiClient
-      : new RDMDepositApiClient(props.config.createUrl, recordSerializer);
+      : new RDMDepositApiClient(apiHeaders, props.config.createUrl, recordSerializer);
```
ntarocco, line 45:
```suggestion
      : new RDMDepositApiClient({headers: apiHeaders}, props.config.createUrl, recordSerializer);
```

ntarocco (approved):
LGTM! Thank you!
@kpsherva @zzacharo if you are happy with this, we can merge and release!
@philippgualdi do you need this backported it to any release?

philippgualdi:
> LGTM! Thank you! @kpsherva @zzacharo if you are happy with this, we can merge and release! @philippgualdi do you need this backported it to any release?

Backported to an older release is not necessary. I appreciate your quick response, thank you! :)

kpsherva (approved):

zzacharo (approved):

PR #589, opened 2022-10-27
Axios: Centralize imports

Pineirin:
closes: inveniosoftware/react-invenio-forms#168


Pineirin:
For reviewers: I was tempted to also move the use of axios in: https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositApiClient.js but decided not to in the end because:

1 - It's the only time we use this axios configuration.
2 - If centralized, the naming could be a bit problematic.

kpsherva (approved):

PR #612, opened 2023-01-18
serializer: send always a pids key to the backend

zzacharo:


slint (approved):
LGTM! Now that I see that you needed to change the test case, I'm curious how the backend PR didn't break anything :thinking:

Issue #611, opened 2023-01-16
Special characters are not correctly escaped


jrcastro2:
## Describe the bug
Special characters are not escpaed in the "submit for review" message.

## Steps to Reproduce
1. Go to 'New upload'.
2. Fill the required data and select a community with sepcial characters in the title.
3. Submit the record for review
4. See the issue.

## Expected behavior
Special characters should be escaped.

## Screenshots (if applicable)
![image](https://user-images.githubusercontent.com/15194802/212635809-8d9a006a-f8a0-4f23-9751-44cd36ee597e.png)

PR #615, opened 2023-02-17
Deposit form: allow special character in submit panel

Pineirin:
closes: #611
Screenshot from 2023-02-17 14-29-49

### DIFF
--- a/src/lib/components/PublishButton/SubmitReviewModal.js
+++ b/src/lib/components/PublishButton/SubmitReviewModal.js
@@ -78,6 +78,7 @@ export class SubmitReviewModal extends Component {
                             communityTitle,
                           }}
                           components={{ bold: <b /> }}
+                          shouldUnescape
                         />
                       }
                       checked={_get(values, "acceptAccessToRecord") === "checked"}


Pineirin:
For reviewers: I would apply this change in more places, as I don't see why <Trans> is needed, but maybe there was a reason behind using it, so I prefer to see first if the change is well received.

src/lib/components/PublishButton/SubmitReviewModal.js
```diff
@@ -70,14 +70,17 @@ export class SubmitReviewModal extends Component {
...
                       label={
-                        <Trans
-                          defaults="The '{{communityTitle}}' curators will have access to
-                 <bold>view</bold> and <bold>edit</bold> your
-                 upload's metadata and files."
-                          values={{
-                            communityTitle,
+                        <div
+                          dangerouslySetInnerHTML={{
+                            __html: i18next.t(
```
alejandromumo, line 75: I am not sure how translations are done and the difference between methods (e.g. using the component `<Trans>` versus `i18next.t()`), so this might be a dummy question: is this change equivalent in terms of translations?

alejandromumo:
> For reviewers: I would apply this change in more places, as I don't see why is needed, but maybe there was a reason behind using it, so I prefer to see first if the change is well received.

I am not sure what you mean by "_I don't see why is needed_".

src/lib/components/PublishButton/SubmitReviewModal.js
Pineirin: From the Trans component documentation:

_While <Trans> gives you a lot of power by letting you interpolate or translate complex React elements, the truth is: in most cases you don't even need it.

As long you have no React/HTML nodes integrated into a cohesive sentence (text formatting like strong, em, link components, maybe others), you won't need it - most of the times you will be using the good old t function._

It is true that in this case we are using bold in the component, which may lead as to think that Trans should be used in this case. Still, as we are needing to dangerouslySetInnerHTML in order to display the special characters, we don't need to use Trans.

Pineirin:
In the end I went with using a property of Trans, instead of setting the inner html.

ppanero (approved):

PR #620, opened 2023-02-27
contrib: added imprint custom field.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1215

### DIFF
--- a/src/lib/components/Imprint.js
+++ b/src/lib/components/Imprint.js
@@ -1,0 +1,75 @@
+import React, { Component } from "react";
+
+import { GroupField, Input } from "react-invenio-forms";
+import { Grid } from "semantic-ui-react";
+
+import PropTypes from "prop-types";
+
+export class Imprint extends Component {
+  render() {
+    const {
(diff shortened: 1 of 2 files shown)


src/lib/components/Imprint.js
```diff
@@ -0,0 +1,76 @@
...
+  render() {
+    const {
+      fieldPath, // injected by the custom field loader via the `field` config property
+      title,
+      publisher,
+      place,
+      isbn,
+      pages,
+    } = this.props;
+    const fieldPathPrefix = `${fieldPath}`;
```
ppanero, line 18: minor: Are we modifying `fieldPath` anywhere? why do we need to create a new variable?

ppanero (approved):

src/lib/components/Imprint.js
alejandromumo: We are not modifying it. 

It's a leftover from the docs, where we are indeed modifying it. So it can be removed I suppose.

PR #628, opened 2023-03-09
community: refactor components to match the invenio_community

kpsherva:
* this is a temporary change, in a long term we won't duplicate these components

* closes https://github.com/inveniosoftware/invenio-communities/issues/912

### DIFF
--- a/src/lib/components/Communities/search/CommunityCompactItemComputer.js
+++ b/src/lib/components/Communities/search/CommunityCompactItemComputer.js
@@ -1,0 +1,88 @@
+// This file is part of InvenioRDM
+// Copyright (C) 2023 CERN.
+//
+// InvenioRDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import { CommunityTypeLabel } from "../labels";
+import _truncate from "lodash/truncate";
+import React from "react";
+import PropTypes from "prop-types";
+
+import { Item, Image, Grid, Icon } from "semantic-ui-react";
+import { RestrictedLabel } from "../labels";
+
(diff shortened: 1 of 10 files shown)


src/lib/components/Communities/search/CommunityCompactItemComputer.js
```diff
@@ -0,0 +1,88 @@
...
+              <RestrictedLabel access={access.visibility} />
+              <CommunityTypeLabel type={communityType} />
+              {extraLabels}
+            </Item.Extra>
+          </Item.Content>
+        </Grid.Column>
+        <Grid.Column width={4}>
+          <Item.Content>
+            <Item.Meta>
+              {ui.permissions.can_direct_publish && (
```
ntarocco, line 60: warning: I renamed this action in my PRs, we will have to pay attention on who merges first :)

src/lib/components/CommunitySelectionModal/CommunityListItem.js
```diff
@@ -5,80 +5,44 @@
...
-        fallbackSrc="/static/images/square-placeholder.png"
-        as={Item.Image}
-      />
+  const actions = (
+    <Button
+      content={itemSelected ? i18next.t("Selected") : i18next.t("Select")}
+      size="small"
+      positive={itemSelected}
+      onClick={() => setLocalCommunity(result)}
+      aria-label={i18next.t("Select ") + metadata.title}
```
ntarocco, line 29: I think this translation in not 100% correct: the variable should be inside the translation, because some languages might write the `Select` after the variable, or in another way :)

jrcastro2 (approved):

PR #631, opened 2023-03-15
Add overradible community modal

jrcastro2:
* Add configurable appName for RSK
* Add configurable options to selection modal
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2076

### DIFF
--- a/src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
+++ b/src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
@@ -43,7 +43,6 @@ export class CommunitySelectionSearch extends Component {
   render() {
     const { selectedEndpoint } = this.state;
     const { allCommunities, myCommunities } = this.apiEndpoints;
-
     const searchApi = new InvenioSearchApi({
       axios: {
         url: selectedEndpoint,
@@ -55,7 +54,6 @@ export class CommunitySelectionSearch extends Component {
       selectedEndpoint === allCommunities
         ? i18next.t("Search in all communities")
         : i18next.t("Search in my communities");
-
     return (
       <OverridableContext.Provider value={overriddenComponents}>
         <ReactSearchKit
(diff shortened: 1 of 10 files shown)


src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
```diff
@@ -37,13 +35,27 @@ export class CommunitySelectionSearch extends Component {
 
     this.state = {
       selectedEndpoint: defaultEndpoint,
+      overriddenComponents: props.overriddenComponents,
     };
   }
 
+  componentDidMount() {
+    const { appName } = this.props;
```
kpsherva, line 43: why was this moved inside the didMount? It most likely causes the components to call the render function twice

src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
```diff
@@ -37,13 +35,27 @@ export class CommunitySelectionSearch extends Component {
...
+    const { overriddenComponents } = this.state;
+    if (_isEmpty(overriddenComponents)) {
+      const defaultOverriddenComponents = {
+        [`${appName}.ResultsList.item`]: CommunityListItem,
+      };
+      this.setState({
+        overriddenComponents: defaultOverriddenComponents,
```
kpsherva, line 50: why we need overriddenComponents inside the state?

src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
jrcastro2: It will render twice only if we pass cmps to override the existing one, but you are right we can move this logic to the constructor itself :+1:

src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js
jrcastro2: Because the cmp overrides cmps for RSK itself, see here: https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/CommunitySelectionModal/CommunitySelectionSearch.js#L60

Initially we want to tune this and override this cmps to display the items differently, as in the mockups there were differences between the modal to select communities in the landing page and the modal from the deposit form, however we decided the keep them the same in the demo, so we could actually remove this.

PR #635, opened 2023-03-27
community: allow search filters on modal

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2076

### DIFF
--- a/src/lib/components/Communities/search/CommunityCompactItemComputer.js
+++ b/src/lib/components/Communities/search/CommunityCompactItemComputer.js
@@ -29,7 +29,7 @@ export const CommunityCompactItemComputer = ({
     >
       <Image as={Item.Image} size="tiny" src={links.logo} />
       <Grid>
-        <Grid.Column width={10}>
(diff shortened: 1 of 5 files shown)


src/lib/components/Communities/search/CommunityCompactItemComputer.js
```diff
@@ -39,19 +39,18 @@ export const CommunityCompactItemComputer = ({
                 {metadata.title}
               </a>
             </Item.Header>
-
+            <div className="mb-5">
```
kpsherva, line 42: why was it moved here? I refactored it lately to follow the correct samntic ui implementation using Item.Extra

jrcastro2: I did this because if we keep the labels in the `Item.Extra` they overlap with the description. 
## Before
Screenshot from 2023-03-29 15-55-00

## Now
Screenshot from 2023-03-29 15-53-54

